PR #3, opened 2020-02-09
Naive sin example

saschatimme:
Trying to get the naive sin example working. Somewhere is a bug or two still..
Also introducing a `@libcall` macro which I think is handy.

### DIFF
--- a/src/arb.jl
+++ b/src/arb.jl
@@ -6,6 +6,11 @@ function Base.string(x::Arb)
     unsafe_string(s)
 end
 
+# Not working, where is c stdout going to???
+# function printn(x::Arb, n::Int, e::Int)
+#     ccall(@libarb(arb_printn), Cvoid, (Ref{Arb}, Int, Int), x, n, e)
+# end
+
 Base.show(io::IO, x::Arb) = print(io, string(x))
 
 @libcall arb_mul_2exp_si 2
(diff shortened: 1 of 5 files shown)


saschatimme:
Things are now computed, but somewhere has to be bug since the output doesn't match with the arb docu http://arblib.org/using.html#a-worked-example-the-sine-function

src/arb.jl
```diff
@@ -0,0 +1,39 @@
+digits_prec(prec::Integer) = floor(Int, prec * log(2) / log(10))
+
+function Base.string(x::Arb)
+    n = digits_prec(x.prec)
+    s = ccall(@libarb(arb_get_str), Cstring, (Ref{Arb}, Int), x, n)
+    unsafe_string(s)
+end
+
+# Not working, where is c stdout going to???
+# function printn(x::Arb, n::Int, e::Int)
```
kalmarek:
```
function printn(x::Arb, n::Int, e::UInt=UInt(0))
    Base.flush(stdout)
    redirect_stdout(stdout) do
        ccall(Arblib.@libarb(arb_printn), Cvoid, (Ref{Arb}, Int, Int), x, n, e)
        Base.Libc.flush_cstdio()
    end
end
```
or something like this

PR #5, opened 2020-02-18
Show

kalmarek:
+ some predicates

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -41,6 +41,12 @@ for (T, args) in (
     end
 end
 
+function set!(res::Arb, str::AbstractString)
+    flag = ccall(@libarb(arb_set_str), Cint, (Ref{Arb}, Cstring, Int), res, str, res.prec)
+    iszero(flag) || throw(ArgumentError("arblib could not parse $str as an Arb"))
+    return res
+end
+
 for (jlT, cT, suffix) in (
     (Arb, Ref{Arb}, :arb_arb),
     (Float64, Cdouble, :d_d),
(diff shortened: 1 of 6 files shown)


src/predicates.jl
```diff
@@ -46,3 +46,66 @@ for (T, funcpairs) in (
...
+            res = ccall(@libarb($arbf), Cint, (Ref{Arf}, $ctype), x, y)
+            return res < 0
+        end
+        Base.:(<)(x::Arf, y::$jltype) = isless(x, y)
+        Base.:(<=)(x::Arf, y::$jltype) = isequal(x,y) | isless(x,y)
+    end
+end
+
+for (ArbT, args) in (
+    (Arb, (
```
saschatimme: Why not also for Arf?

src/constructors.jl
```diff
@@ -75,26 +81,32 @@ for ArbT in (:Arf, :Arb, :Acb, :Mag)
...
 
-Arf(x::BigFloat, prec::Integer) = set!(Arf(prec), x)
-Arb(x::BigFloat, prec::Integer) = set!(Arb(prec), Arf(x, prec))
-Acb(x::BigFloat, prec::Integer) = set!(Acb(prec), Arb(x, prec))
+Arf(x::BigFloat, prec::Integer=precision(x)) = set!(Arf(prec), x)
+Arb(x::BigFloat, prec::Integer=precision(x)) = set!(Arb(prec), Arf(x, prec))
+Acb(x::BigFloat, prec::Integer=precision(x)) = set!(Acb(prec), Arb(x, prec))
 
 # fallbacks:
 Arf(x::Real, prec::Integer) = set!(Arf(prec), BigFloat(x))
```
saschatimme: I think we should still have a global `precision` variable as a fallback. Just if you want to be lazy and just quickly try something.

src/predicates.jl
kalmarek: there are no `arf_eq`, etc. `libarb` implements only `arf_equal` and `arf_cmp*`

src/constructors.jl
kalmarek: ok, but what about thread safeness?

saschatimme: Sure, this is not sensible for thread safe code, but you also do not write hastily thread safe code 

I am just thinking of defining
```
const DEFAULT_PREC = Ref(256)
precision(x::Number) = DEFAULT_PREC[]
```

In library code, one definitely should always carry around the precision

PR #6, opened 2020-02-19
Sin naive2

kalmarek:


saschatimme:
I rebased onto master, fixed some errors introduces by moving precision to a kwarg and added the example also as a test case

PR #7, opened 2020-02-19
Add gh actions ci

saschatimme:


saschatimme:
No need to go crazy with tests, but let's make sure we keep a working master :)

PR #11, opened 2020-04-12
Arbcall tests

Joel-Dahne:
To get a feeling for the package I though it would be a good idea to add some tests. I have added tests for parts of the `arbcall` functionality. More precisely for testing `jlfname`, `returntype` and `Carg`, as you can see from the commits.

In addition to just testing the intended behavior of the code I have added test cases which highlights possible changes or improvements. For example there are two tests for `jlfname` which test
```julia
@test jlfname("mag_set_d_lower") == :set_d_lower
@test jlfname("arb_ui_div") == :ui_div)
```
Both which are candidates for changing in the future.

It's written in such a way that all the tests passes at the moment. Not matter if it might be the intended behavior or not.

### DIFF
--- a/test/arbcall-test.jl
+++ b/test/arbcall-test.jl
@@ -1,0 +1,46 @@
+@testset "jlfname" begin
+    for (arbfname, name) in (# Supported types
+                             ("mag_set", :set),
+                             ("mag_set_d", :set),
+                             ("mag_set_ui", :set),
+                             ("arf_set", :set),
+                             ("arf_set_ui", :set),
+                             ("arf_set_si", :set),
+                             ("arf_set_d", :set),
+                             ("arb_set", :set),
+                             ("arb_set_arf", :set),
+                             ("arb_set_si", :set),
+                             ("arb_set_ui", :set),
+                             ("arb_set_d", :set),
+                             ("acb_set", :set),
+                             ("acb_set_ui", :set),
+                             ("acb_set_si", :set),
+                             ("acb_set_d", :set),
+                             ("acb_set_arb", :set),
+                             ("acb_set_si_si", :set),
+                             ("acb_set_d_d", :set),
+                             ("acb_set_arb_arb", :set),
+
+                             # Unsupported types
+                             ("arf_set_fmpz", :set_fmpz),
+                             ("arf_set_mpfr", :set_mpfr),
+                             ("acb_set_fmpq", :set_fmpq),
+                             ("arb_bin_uiui", :bin_uiui),
+
+                             # Deprecated types
+                             ("arf_set_mpz", :set_mpz),
+                             ("arf_set_fmpr", :set_fmpr),
+
+                             # Underscore methods
+                             ("_arb_vec_set", :_arb_vec_set),
+
+                             # Some special cases to be aware of and maybe change
+                             ("mag_set_d_lower", :set_d_lower),
+                             ("arb_ui_div", :ui_div),
+                             ("arb_rising_ui_rec", :rising_ui_rec),
+                             ("arb_zeta_ui_vec", :zeta_ui_vec),
+                             )
+        @test Arblib.jlfname(arbfname) == name
+        @test Arblib.jlfname(arbfname, inplace = true) == Symbol(name, :!)
+    end
+end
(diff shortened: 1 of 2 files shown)


kalmarek:
that's cool, thanks so much! I will try to read through this tomorrow.
If i understand your intention lets discuss first if these are the behaviours (`jlfnames`) we want in the long term

saschatimme:
Closed and reopened to get CI running :)

Joel-Dahne:
My intentions were more to document the current behaviour, highlighting things that are candidates for changing. The discussion about what behaviour to prefer could be had here, or at some later point and updating the tests then.

test/arbcall-test.jl
```diff
@@ -0,0 +1,130 @@
...
+                             ("acb_set_si", :set),
+                             ("acb_set_d", :set),
+                             ("acb_set_arb", :set),
+                             ("acb_set_si_si", :set),
+                             ("acb_set_d_d", :set),
+                             ("acb_set_arb_arb", :set),
+
+                             # Unsupported types
+                             ("arf_set_fmpz", :set_fmpz),
+                             ("arf_set_mpfr", :set_mpfr),
```
saschatimme, line 87: I think we should be able to support MPFR since these are just `BigFloat`s.

test/arbcall-test.jl
```diff
@@ -0,0 +1,130 @@
...
+                             ("arf_set_fmpr", :set_fmpr),
+
+                             # Underscore methods
+                             ("_arb_vec_set", :_arb_vec_set),
+
+                             # Some special cases to be aware of and maybe change
+                             ("mag_set_d_lower", :set_d_lower),
+                             ("arb_ui_div", :ui_div),
+                             ("arb_rising_ui_rec", :rising_ui_rec),
+                             ("arb_zeta_ui_vec", :zeta_ui_vec),
```
saschatimme, line 102: Right now, we only remove `ui` etc. if it appears at the end. However, we could also consider removing it at any position, i.e., `arb_zeta_ui_vec` -> `zeta_vec`.

test/arbcall-test.jl
Joel-Dahne: Yes, this should be a simple change. I also think it could be nice to support `fmpz` by creating a very simple type for it which only purpose is to convert between `BigInt` and `fmpz`, but I think it is better to that leave for later.

test/arbcall-test.jl
Joel-Dahne: I think this is reasonable. The types should still all be different to there should not be any clashing name issues.

test/arbcall-test.jl
```diff
@@ -0,0 +1,130 @@
...
+                             # Deprecated types
+                             ("arf_set_mpz", :set_mpz),
+                             ("arf_set_fmpr", :set_fmpr),
+
+                             # Underscore methods
+                             ("_arb_vec_set", :_arb_vec_set),
+
+                             # Some special cases to be aware of and maybe change
+                             ("mag_set_d_lower", :set_d_lower),
+                             ("arb_ui_div", :ui_div),
```
kalmarek, line 100: I'd have thought that this function would be named `arb_div_ui_arb`, but anyway:
We can either add `ui` to the list of prefixes, or keep it like it is.

When designing signature parsing we planned to allow providing julia function name, when heuristic for parsing the signature doesn't work: instead of using
`arbcall"void arb_ui_div(arb_t z, ulong x, const arb_t y, slong prec)"`
we'd need to use e.g.
```julia
@eval begin
   af = ArbFuncion("void arb_ui_div(arb_t z, ulong x, const arb_t y, slong prec)")
   jlcode(af, :div)
end
```
or  add 
```julia
macro arbcall(jlf_name::Symbol, signature_str)
    af = ArbFunction(signature_str)
    return jlcode(af, jlf_name)
end
```
for this purpose.

test/arbcall-test.jl
kalmarek: I think some hand-written `set!` already use BigFloat:
https://github.com/kalmarek/Arblib.jl/blob/981f097003d484d86be2f6f442e105e4818eb673/src/constructors.jl#L7
so let's just add it to the "supported" list

test/arbcall-test.jl
kalmarek: let's do it then

Joel-Dahne:
Great! Then there are a number of possible improvements in the future! 

I have already started working on some more things related to testing the signatures of the functions, https://github.com/Joel-Dahne/Arblib.jl/tree/arbcall-signature. It's work on top of this branch so I think it's easiest if I create a pull request for that after this pull request has been merged (from my understanding of how GitHub handles this at least).

PR #13, opened 2020-04-14
Arbcall signatures

Joel-Dahne:
This pull request contains two rather harmless commits and one more opinionated. The first two commits adds methods for computing the signatures of `Arbfunctions`, one for Julia signature and one for the arb signature. The Julia signature is not to interesting, it mainly makes it easier to test this part separately (it's slightly different from the previously used signature, more about that later).

The arb signature is more interesting, it allows partially automatic testing of `Arbfunction`. I have used this to do some testing. I began by extracting all the functions from the documentation of arb
```sh
grep function:: doc/source/*.rst | cut -f 3- -d " " > $SOMEPATH
```
I then wrote some Julia code to parse all of these functions and compare the arb signature of the result with the original string.
```julia
function parse_arb(filename)
    open(filename) do f
        methods = Arbfunction[]
        underscoremethods = Arbfunction[]
        type_error = String[]
        parse_error = String[]
        for line in readlines(f)
            try
                f = Arbfunction(line)

                if(line != arbsignature(f))
                    println(line)
                    println(arbsignature(f))
                end

                if startswith(string(jlfname(f)), "_")
                    push!(underscoremethods, f)

                else
                    push!(methods, f)
                end
            catch e
                if e isa KeyError
                    push!(type_error, line)
                elseif e isa ArgumentError
                    push!(parse_error, line)
                else
                    rethrow(e)
                end
            end
        end

        println("$(length(methods)) correctly parsed methods")
        println("$(length(underscoremethods)) correctly parsed underscore methods")

        println("$(length(type_error)) methods that were not parsed due to unsupported type")
        println("$(length(parse_error)) methods for which parsing failed")

        (methods, underscoremethods, type_error, parse_error)
    end
end
```
Running it on the above generated file give
```
julia> (m, u, t, p) = Arblib.parse_arb("allall.txt");
752 correctly parsed methods
6 correctly parsed underscore methods
1397 methods that were not parsed due to unsupported type
68 methods for which parsing failed
```
This was done using the master branch of arb (the last commit fixes a typo in the documentation). I have not included code for this in the commits, mainly because I was not sure how to handle the arb documentation files in a good way. Most of the fails are due to unsupported types, this will be fixed as more types are added (for example polynomials and matrices). The methods for which the parsing fails all contain pointers, for example `int acb_mat_lu_classical(slong * perm, acb_mat_t LU, const acb_mat_t A, slong prec)`. There is one exception, but this is just a typo in the documentation (I'll probably submit a pull request to arb for this).

Finally to the opinionated commit. Previously the code generated by `jlcode` always returned the first argument. This follows the Julia default for in-place methods but doesn't work well for other types of methods. For example it doesn't make sense for predicates, where the return value is the important part. You could consider returning the first argument for void functions, however this will be inconsistent due to the fact that the `arf` functions return an `int` and the `arb` functions return `void`. You could also consider returning the first argument for all in-place methods, but it might be that some of them have return values which are important (I have not checked).

The change I made know makes the methods always return the return value from arb. I would say this is the safest choice, no information is lost, but it goes against the Julia default. It's also nice because the signature of the Julia and arb versions then match. I don't know what the best choice is.

### DIFF
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -10,6 +10,8 @@ const Ctypes = Dict{String, DataType}(
     "mag_t"     => Mag,
     "arf_rnd_t" => arb_rnd,
     "mpfr_t"    => BigFloat,
+    "char *"    => Cstring,
+    "slong *"   => Vector{Clong},
 )
 
 struct Carg{ArgT}
@@ -18,7 +20,7 @@ struct Carg{ArgT}
 end
 
 function Carg(str)
-    m = match(r"(?<const>const)?\s*(?<type>\w+)\s+(?<name>\w+)", str)
+    m = match(r"(?<const>const)?\s*(?<type>\w+(\s\*)?)\s+(?<name>\w+)", str)
     isnothing(m) && throw(ArgumentError("string doesn't match c-argument pattern"))
 
     return Carg{Ctypes[m[:type]]}(m[:name], !isnothing(m[:const]))
@@ -29,6 +31,7 @@ isconst(ca::Carg) = ca.isconst
 jltype(::Carg{ArgT}) where ArgT = ArgT
 ctype(ca::Carg) = jltype(ca)
 ctype(::Carg{ArgT}) where ArgT <: Union{Arf, Arb, Acb, Mag, BigFloat}  = Ref{ArgT}
+ctype(::Carg{Vector{T}}) where T = Ref{T}
 
 struct Arbfunction{ReturnT}
     fname::String
@@ -36,7 +39,7 @@ struct Arbfunction{ReturnT}
 end
 
 function Arbfunction(str)
-    m = match(r"(?<returntype>\w+)\s+(?<arbfunction>[\w_]+)\((?<args>.*)\)",
+    m = match(r"(?<returntype>\w+(\s\*)?)\s+(?<arbfunction>[\w_]+)\((?<args>.*)\)",
         str)
     isnothing(m) && throw(ArgumentError("string doesn't match arblib function signature pattern"))
 
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -132,47 +132,14 @@ end
 function jlcode(af::Arbfunction, jl_fname=jlfname(af))
     returnT = returntype(af)
     args = arguments(af)
-    c_types = ctype.(args)
 
-    arg_names = Symbol.(name.(args))
-    jl_types = jltype.(args)
-
-    kwargs = Expr[]
-
-    k = findfirst(==(:prec), arg_names)
-    if !isnothing(k)
-        @assert jl_types[k] == Int64
-        p = :prec
-        a = first(args)
-        default = if jltype(a) ∈ (Arf, Arb, Acb)
-            :(precision($(Symbol(name(a)))))
-        else
-            :(DEFAULT_PRECISION[])
-        end
-        push!(kwargs, Expr(:kw, :($p::Integer), default))
-        deleteat!(arg_names, k)
-        deleteat!(jl_types, k)
-    end
-
-    k = findfirst(==(:rnd), arg_names)
-    if !isnothing(k)
-        @assert jl_types[k] == arb_rnd
-        r = :rnd
-        push!(kwargs, Expr(:kw, :($r::Union{arb_rnd, RoundingMode}), :(RoundNearest)))
-        deleteat!(arg_names, k)
-        deleteat!(jl_types, k)
-    end
-
-    jl_args = [:($a::$T) for (a, T) in zip(arg_names, jl_types)]
-
-    return :(
-        function $jl_fname($(jl_args...); $(kwargs...))::$returnT
-        ccall(Arblib.@libarb($(arbfname(af))),
-              $returnT,
-              $(Expr(:tuple, c_types...)),
-              $(Symbol.(name.(args))...))
-        end
-    )
+    return :($(jlsignature(af)) = begin
+             ccall(Arblib.@libarb($(arbfname(af))),
+                   $returnT,
+                   $(Expr(:tuple, ctype.(args)...)),
+                   $(Symbol.(name.(args))...))
+             end
+             )
 end
 
 macro arbcall_str(str)
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -8,7 +8,8 @@ const Ctypes = Dict{String, DataType}(
     "arb_t"     => Arb,
     "acb_t"     => Acb,
     "mag_t"     => Mag,
-    "arf_rnd_t" => arb_rnd
+    "arf_rnd_t" => arb_rnd,
+    "mpfr_t"    => BigFloat,
 )
 
 struct Carg{ArgT}
@@ -27,7 +28,7 @@ name(ca::Carg) = ca.name
 isconst(ca::Carg) = ca.isconst
 jltype(::Carg{ArgT}) where ArgT = ArgT
 ctype(ca::Carg) = jltype(ca)
-ctype(::Carg{ArgT}) where ArgT <: Union{Arf, Arb, Acb, Mag}  = Ref{ArgT}
+ctype(::Carg{ArgT}) where ArgT <: Union{Arf, Arb, Acb, Mag, BigFloat}  = Ref{ArgT}
 
 struct Arbfunction{ReturnT}
     fname::String
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -75,26 +75,24 @@ function jlfname(af::Arbfunction,
     return jlfname(arbfname(af), prefixes, suffixes, inplace=inplace)
 end
 
-function jlsignature(af::Arbfunction)
-    returnT = returntype(af)
-    args = arguments(af)
+function jlargs(af::Arbfunction)
+    cargs = arguments(af)
+    arg_names = Symbol.(name.(cargs))
+    jl_types = jltype.(cargs)
 
-    arg_names = Symbol.(name.(args))
-    jl_types = jltype.(args)
-
-    jl_kwargs = Expr[]
+    kwargs = Expr[]
 
     k = findfirst(==(:prec), arg_names)
     if !isnothing(k)
         @assert jl_types[k] == Int64
         p = :prec
-        a = first(args)
+        a = first(cargs)
         default = if jltype(a) ∈ (Arf, Arb, Acb)
             :(precision($(Symbol(name(a)))))
         else
             :(DEFAULT_PRECISION[])
         end
-        push!(jl_kwargs, Expr(:kw, :($p::Integer), default))
+        push!(kwargs, Expr(:kw, :($p::Integer), default))
         deleteat!(arg_names, k)
         deleteat!(jl_types, k)
     end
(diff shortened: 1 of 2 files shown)


Joel-Dahne:
You were faster at replying than I were at writing the first comment. I think your points are the same as the one I makes as well. I think returning the modified argument is the best choice in general. Here I were conservative and stayed as close to arb as possible.

kalmarek:
@Joel-Dahne no, it's just github automatically pulled my comment from one of the commits in your fork ;)
Great work btw!

I would opt to warn/error on non-zero return arf functions, but I'd always return the first argument if it's not marked as `const`; but following arb closely has its own advantages. @saschatimme?

Joel-Dahne:
According to the arb documentation we have for arf

> The return flag is 0 if the result is exact (not rounded) and 1 if the result is inexact (rounded)

So warning for non-zero doesn't make much sense, most of the time the result is not exact. Note that MPFR has a similar return value, but `BigFloat` completely ignores this.

saschatimme:
Mhm we could also have an optional positional argument in this case which controls the behaviour where the default is the Julia style of returning the modified argument.

Joel-Dahne:
An example of a mutating function for which the return value is important

> int arb_intersection(arb_t z, const arb_t x, const arb_t y, slong prec)
>    If x and y overlap according to arb_overlaps(), then z is set to a ball containing the intersection of x and y and a nonzero value is returned. Otherwise zero is returned and the value of z is undefined. If x or y contains NaN, the result is NaN.

kalmarek:
I went through and looked into
```
grep function:: doc/source/*.rst | cut -f 3- -d " " | grep -E "(arf_|arb_|acb_)" > all.txt`
```
```julia
(m, u, t, p) = parse_arb("/tmp/arb/all.txt");
661 correctly parsed methods
6 correctly parsed underscore methods
1132 methods that were not parsed due to unsupported type
48 methods for which parsing failed
```
Of those 48 methods:
* 17 contain `FILE` handlers (which need to be dealt manually, anyway I believe),
* adding `(\s\*)?` to regexes for arguments and return types makes this number to just one:
`"partitions_hrr_sum_arb(arb_t x, const fmpz_t n, slong N0, slong N, int use_doubles)"`

When it comes to `slong *` and so if could make into `Ctypes` as `Vector{Clong}`

(this stays the same for the original `allall.txt`)

kalmarek:
I just went searching in Arb documentation for undefined behaviors; It seems the only case for arf/arb/acb functions is `arb_intersection`. however there are quite a few of these for `(arb|acb|)_mat`s and special functions

Joel-Dahne:
> "partitions_hrr_sum_arb(arb_t x, const fmpz_t n, slong N0, slong N, int use_doubles)"
Is a typo in the documentation, it's missing the return type. I can submit a pull request to arb for this later.

kalmarek:
there are also these 6 which don't follow the formatting (no space after `*`)
```
 "char* arf_dump_str(const arf_t x)"
 "int arf_dump_file(FILE* stream, const arf_t x)"
 "int arf_load_file(arf_t x, FILE* stream)"
 "char* mag_dump_str(const mag_t x)"
 "int mag_dump_file(FILE* stream, const mag_t x)"
 "int mag_load_file(mag_t x, FILE* stream)"
```

kalmarek:
also: 
* if a function returns `char *`the returned pointer must be `unsafe_string`ed followed by `ccall((:flint_free, libflint), Cvoid, ...)`
* the functions which need `FILE * stream` argument will need to take `file_name::(Abstract?)String` argument and
```julia
open(file_name, "w") as file
    fstream = Libc.FILE(file)
    ccall(....) # fstream passed as `Ptr{Cvoid}`
    close(fstream)
end
```

I guess both classes can't be handled automatically, unless we write more heuristics for `jlcode`

Joel-Dahne:
The most reasonable thing to do might be to just ignore those cases. We can maybe still make sure it parses correctly but that `Arbfunction` treats it as an unsupported type. These can then be handled as special cases when necessary. I think most of them are convenient  functions that are not too relevant in Julia, for example loading from a file would be done in a different way.

Joel-Dahne:
I believe the implementation in this pull is the most conservative one, we wrap the functions from arb as is without modifying any return values. This is most likely not the best choice, but maybe we can take it for now and make a more informed choice later on. Changing it later on would be fairly simple, so we are not locking ourselves into something suboptimal I believe.

For now I'll try to continue working on the parsing of functions, adding more types, improve the error handling (I want to distinguish between unsupported types (like FILE) and types which are not implemented (like arb_poly)). None of this depends on our choice of return type.

src/arbcall.jl
```diff
@@ -107,17 +165,14 @@ function jlcode(af::Arbfunction, jl_fname=jlfname(af))
...
-            $(Symbol.(name.(args))...))
-            return $res
+        function $jl_fname($(jl_args...); $(kwargs...))::$returnT
+        ccall(Arblib.@libarb($(arbfname(af))),
+              $returnT,
+              $(Expr(:tuple, c_types...)),
+              $(Symbol.(name.(args))...))
         end
-        )
+    )
```
Joel-Dahne, line 175: It would be great if someone can figure out how to use `jlsignature(af)` here. Just
```
return :(
    function $(jlsignature(af))
    ccall(Arblib.@libarb($(arbfname(af))),
          $returnT,
          $(Expr(:tuple, c_types...)),
          $(Symbol.(name.(args))...))
    end
)
```
did not work.

kalmarek: to make it work you may need to modify AST directly;
also note that in `jlsignature` (as defined) `x::Arb` means something different (type assertion) than in `function is_zero(x::Arb) end`.

```julia
julia> f = Arbfunction("int arb_is_zero(const arb_t x)");
julia> dump(Arblib.jlsignature(f))
Expr
  head: Symbol ::
  args: Array{Any}((2,))
    1: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: Symbol is_zero
        2: Expr
          head: Symbol parameters
          args: Array{Any}((0,))
        3: Expr
          head: Symbol ::
          args: Array{Any}((2,))
            1: Symbol x
            2: Arb <: Real
              arb::Arblib.arb_struct
              prec::Int64
    2: Int32 <: Signed

julia> ex = :(function is_zero(x::Arb; )::Int32 end); dump(ex)
Expr
  head: Symbol function
  args: Array{Any}((2,))
    1: Expr
      head: Symbol ::
      args: Array{Any}((2,))
        1: Expr
          head: Symbol call
          args: Array{Any}((3,))
            1: Symbol is_zero
            2: Expr
              head: Symbol parameters
              args: Array{Any}((0,))
            3: Expr
              head: Symbol ::
              args: Array{Any}((2,))
                1: Symbol x
                2: Symbol Arb
        2: Symbol Int32
    2: Expr
      head: Symbol block
      args: Array{Any}((2,))
        1: LineNumberNode
          line: Int64 2
          file: Symbol REPL[7]

```

Joel-Dahne: Yes, I had a lot of issues with this trying to write the tests. But okay, I'll take another look and see if I can modify the AST directly.

Joel-Dahne: I got it to work, it was not to complicated in the end. The new version does not produce identical AST to the old one, but the result should be the same. I also added some very simple tests to make sure you at least can call the generated methods.

src/arbcall.jl
```diff
@@ -100,24 +99,47 @@ function jlcode(af::Arbfunction, jl_fname=jlfname(af))
...
+    c_args = join([ifelse(isconst, "const ", "")*"$type $name" for (isconst, type, name)
+                   in zip(arg_consts, arg_ctypes, arg_names)], ", ")
+
+    c_args = join([ifelse(isconst(arg), "const ", "") *
+                   "$(jltoctype[jltype(arg)]) $(name(arg))" for arg in args], ", ")
+
+    "$creturnT $(arbfname(af))($c_args)"
+end
+
+function jlcode(af::Arbfunction, jl_fname=jlfname(af))
```
kalmarek, line 132: it would be best to keep the external control on what is the function named on julia side;
as is `jl_fname` has no effect on `jlsignature` this control is lost

src/arbcall.jl
```diff
@@ -100,24 +99,47 @@ function jlcode(af::Arbfunction, jl_fname=jlfname(af))
...
-        push!(kwargs, Expr(:kw, :($r::Union{arb_rnd, RoundingMode}), :(RoundNearest)))
+        push!(jl_kwargs, Expr(:kw, :($r::Union{arb_rnd, RoundingMode}), :(RoundNearest)))
         deleteat!(arg_names, k)
         deleteat!(jl_types, k)
     end
 
     jl_args = [:($a::$T) for (a, T) in zip(arg_names, jl_types)]
 
-    res = first(arg_names)
+    :($(jlfname(af))($(jl_args...); $(jl_kwargs...))::$(returnT))
```
kalmarek, line 109: to be honest it is not clear to me what is the actual benefit of `jlsignature` (except maybe testing?)
to make both testing and the generated code reading easier I suggest that the function just returns a tuple
```julia
return (jl_args, jl_kwargs)
```
note that `returnT` and `jlfname` can and should be computed directly from `af::Arbfunction` and not tangled into another function, when the're directly accessible outside it;
maybe renaming the function to `jlargs(af::Arbfunction)` also makes sense?

src/arbcall.jl
```diff
@@ -100,24 +99,47 @@ function jlcode(af::Arbfunction, jl_fname=jlfname(af))
...
+                   in zip(arg_consts, arg_ctypes, arg_names)], ", ")
+
+    c_args = join([ifelse(isconst(arg), "const ", "") *
+                   "$(jltoctype[jltype(arg)]) $(name(arg))" for arg in args], ", ")
+
+    "$creturnT $(arbfname(af))($c_args)"
+end
+
+function jlcode(af::Arbfunction, jl_fname=jlfname(af))
+    returnT = returntype(af)
```
kalmarek, line 133: with the remark on `jlsignature` applied this can be rewritten as
```julia
returnT = returntype(af)
cargs = arguments(af)
args, kwargs = jlargs(af)

return :(function $(jl_fname)($(args...); $(kwargs...))::$(returnT)
    (....) # the rest remains the same
end
```

kalmarek:
@Joel-Dahne sorry, I just noticed that this is your private branch (why did github allowed me to push there?); I hope I didn't interrupted your workflow

src/arbcall.jl
Joel-Dahne: I must admit I mainly wrote it for testing purposes, and then thought it would be natural to use it in `jlcode` as well. But I agree, to much information is mangled into this one method. I can change it to `jlargs` as you propose!

Joel-Dahne:
> @Joel-Dahne sorry, I just noticed that this is your private branch (why did github allowed me to push there?); I hope I didn't interrupted your workflow

No worries, you can choose to allow or not allow commits from reviewers when you create the pull request (default is to allow it I believe)

Joel-Dahne:
I can also note that I submitted I pull request to arb, fixing the documentation for the functions that we noted have issues.

src/arbcall.jl
Joel-Dahne: Now I have rewritten it in terms of `jlargs`. This does feel cleaner!

kalmarek:
I think we can merge this, unless there is something you want to add here?

Joel-Dahne:
I'm fine with merging this now.

PR #15, opened 2020-04-19
relax jltypes

kalmarek:
consider (old behaviour)
```julia
julia> af = Arblib.Arbfunction("void arb_pow_ui(arb_t y, const arb_t b, ulong e, slong prec)")
Arblib.Arbfunction{Nothing}("arb_pow_ui", Arblib.Carg[Arblib.Carg{Arb}("y", false), Arblib.Carg{Arb}("b", true), Arblib.Carg{UInt64}("e", false), Arblib.Carg{Int64}("prec", false)])

julia> Arblib.jlcode(af)
:(function pow!(y::Arb, b::Arb, e::UInt64; prec::Integer = precision(y))
      #= /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:141 =#
      ccall(#= /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:141 =# Arblib.@libarb("arb_pow_ui"), Nothing, (Ref{Arb}, Ref{Arb}, UInt64, Int64), y, b, e, prec)
  end)
```
this defines `pow!` that needs `UInt64` exponent;
however `ccall` calls `Base.cconvert(UInt64, e)` internally, which will result with the correct thing for any `e` that converts to `UInt64`.

I suggest that we relax signatures in `jlargs` (new behaviour):
```julia
julia> af = Arblib.Arbfunction("void arb_pow_ui(arb_t y, const arb_t b, ulong e, slong prec)")
Arblib.Arbfunction{Nothing}("arb_pow_ui", Arblib.Carg[Arblib.Carg{Arb}("y", false), Arblib.Carg{Arb}("b", true), Arblib.Carg{UInt64}("e", false), Arblib.Carg{Int64}("prec", false)])

julia> Arblib.jlcode(af)
:(function pow!(y::Arb, b::Arb, e::Integer; prec::Integer = precision(y))
      #= /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:144 =#
      ccall(#= /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:144 =# Arblib.@libarb("arb_pow_ui"), Nothing, (Ref{Arb}, Ref{Arb}, UInt64, Int64), y, b, e, prec)
  end)

julia> x = Arb();

julia> Arblib.pow!(x, Arb(2.0), big(3))

julia> x
8.0000000000000000000000000000000000000000000000000000000000000000000000000000
```

### DIFF
--- a/src/arb.jl
+++ b/src/arb.jl
@@ -1,9 +1,9 @@
-@libcall arb_mul_2exp_si 2
-@libcall arb_one 1
-@libcall arb_pow_ui 2 prec=true
-@libcall arb_fac_ui 1 prec=true
-@libcall arb_div 3 prec=true
-@libcall arb_abs 2 prec=true
-@libcall arb_add_error 2
-@libcall arb_add 3 prec=true
-@libcall arb_sub 3 prec=true
+arbcall"void arb_mul_2exp_si(arb_t y, const arb_t x, slong e)"
+arbcall"void arb_one(arb_t f)"
+arbcall"void arb_pow_ui(arb_t y, const arb_t b, ulong e, slong prec)"
+arbcall"void arb_fac_ui(arb_t z, ulong n, slong prec)"
+arbcall"void arb_div(arb_t z, const arb_t x, const arb_t y, slong prec)"
+arbcall"void arb_abs(arb_t y, const arb_t x)"
+arbcall"void arb_add_error(arb_t x, const arb_t err)"
+arbcall"void arb_add(arb_t z, const arb_t x, const arb_t y, slong prec)"
+arbcall"void arb_sub(arb_t z, const arb_t x, const arb_t y, slong prec)"
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -32,9 +32,12 @@ isconst(ca::Carg) = ca.isconst
 rawtype(::Carg{T}) where T = T
 
 jltype(ca::Carg) = rawtype(ca)
-jltype(::Carg{<:AbstractFloat}) = AbstractFloat
-jltype(::Carg{<:Integer}) = Integer
-jltype(::Carg{Cstring}) = AbstractString
+jltype(ca::Carg{Cint}) = Integer
+jltype(ca::Carg{Clong}) = Integer
+jltype(ca::Carg{Culong}) = Unsigned
+jltype(ca::Carg{Cdouble}) = Base.GMP.CdoubleMax
+jltype(ca::Carg{arb_rnd}) = Union{arb_rnd, RoundingMode}
+jltype(ca::Carg{Vector{Clong}}) = Vector{<:Integer}
 
 ctype(ca::Carg) = rawtype(ca)
 ctype(::Carg{T}) where T <: Union{Arf, Arb, Acb, Mag, BigFloat}  = Ref{T}
@@ -85,13 +88,14 @@ end
 function jlargs(af::Arbfunction)
     cargs = arguments(af)
     arg_names = Symbol.(name.(cargs))
+    c_types = ctype.(cargs)
     jl_types = jltype.(cargs)
 
     kwargs = Expr[]
 
     k = findfirst(==(:prec), arg_names)
     if !isnothing(k)
-        @assert jl_types[k] == Integer
+        @assert c_types[k] == Clong
         p = :prec
         a = first(cargs)
         default = if jltype(a) ∈ (Arf, Arb, Acb)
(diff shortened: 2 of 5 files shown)


Joel-Dahne:
I think this is in the spirit of how Julia normally handles type conversions for numbers and I think it should be done. The only problem I see is when arb has both `slong` and `ulong` versions of a function, they would then have identical signatures in Julia. One could consider specializing into `Signed` and `Unsigned` in this case, but then they would need to know about each other somehow.

Joel-Dahne:
I think they case of `Bool` versus `Int32` is interesting as well. While many of the `int` types in arb represents booleans, some of them are flags with more possible values. It would fit much better in Julia to return a `Bool` instead of an `Int32` when possible, but then we need some way of determining when this can be done. When it occurs as an argument we can use `Union{Bool, Int32}` or maybe even `Integer`.

kalmarek:
That's a good point. In such cases, we'd need to pick one, default version, and maybe write the other by hand? `Signed` and `Unsigned` sound much better, but that is a solution only when both are present. Any ideas?

As for return types I'd go for trait dispatched macro:
```julia
@arbcall Predicate "int arb_is_zero(const arb_t x)" iszero
```
but that's a topic for separate issue.

Joel-Dahne:
Something which we can do completely safely is to use `ClongMax` and `CulongMax` defined in `Base.GMP` for `slong` and `ulong` respectively, they are just unions of integers of sizes 16, 32 and 64. 

Maybe we can default to `Integer` for both and then make `ulong` into `Unsigned` by hand when required? We can maybe check how many cases there are of this in arb? I plan to write some more code for parsing which would make these types of things easier to check.

To make things even more complicated we also have `fmpz`, for example `BigInt` would naturally be converted to that. Then the natural thing would probably be to have `ClongMax` and `CulongMax` for `slong` and `ulong` and `Integer` for `fmpz`.

kalmarek:
@Joel-Dahne 
> To make things even more complicated we also have fmpz, for example BigInt would naturally be converted to that. Then the natural thing would probably be to have ClongMax and CulongMax for slong and ulong and Integer for fmpz.

I think it's ok to have one default version for `Integer` dispatching to `arb_*_si` and then separate for
* `Unsigned` → `arb_*_ui`
* `BigInt` → `fmpz`
written by hand. I don't have time right now to see how much of these would need to be written by hand, but I assume quite a lot of simple arithmetic would. One possible solution could be to
```
@arbcall "void arb_add_ui(arb_t y, const arb_t b, ulong e)" "ulong"=> Unsigned
```
with type overrides stored in additional field of `Arbfunction`. then we could modify `jlargs` on the fly

Joel-Dahne:
I like the idea of being able to specify it in a simple way. I would propose to do it based on the name of the argument though, to avoid problems when several arguments have the same type.
```
@arbcall "void arb_add_ui(arb_t y, const arb_t b, ulong e)" "e"=> Unsigned
```
If it's very easy to generate these methods on the fly then the end user could even generate them as needed. We can supply a default version of most methods but if they need more or just different version they can handle that themselves. Though I'm not sure how Julia treats this in general, it might give weird consequences in some cases.

Joel-Dahne:
The types we currently handle are `Cvoid, Cint, Clong, Culong,
Cdouble, Arf, Arb, Acb, Mag, arb_rnd, BigFloat, Cstring` and
`Vector{Clong}`. I would propose that the following types stay as they
are

- Arf
- Arb
- Mag
- Acb
- BigFloat
- Cvoid
- Cstring 

The first four are somewhat obvious. For `Cvoid` and `Cstring` there
is basically nothing that converts to them anyway, in particular
`String` doesn't convert to `Cstring` automatically. `BigFloat` is
related to a comment further down about automatically converting
between different precisions.

For the other ones I would propose

- `Cint` -> `Integer`
- `Clong` -> `Integer`
- `Culong` -> `Unsigned`
- `Cdouble` -> `Base.GMP.CdoubleMax`
- `arb_rnd` -> `Union{arb_rnd, RoundingMode}`
- `Vector{Clong}` -> `Vector{<:Integer}`

For `arb_rnd` you have already defined this conversion, so it should
be safe. All of the integer conversions are safe in the sense that if
you give a number that is to large to be represented it throws and
error. However, as we have discussed, many functions have both an `si`
and a `ui` version so differentiating between those makes it so that
the signature do not clash. We would still have that a function which
works with both `Cint` and `Clong` would have clashing signarues,
however I don't think there are any of them, or at least very few.
When there are functions that only accept `ui` we can then maybe
manually change it to `Integer` instead.

I would propose to be very conservative with automatically converting
floating points, I hate when methods silently throw away my precious
precision. For integers the conversion is safe, if it can be
represented it is exact, but for floats it's not. For example if I write

```
x = Arb(1)
y = big"1.1"
z = Arb(0)
add!(z, x, y)
```

I expect the full precision to work. You can get very annoying bugs if
this silently truncates `y` to 53 bits of precision. I have shot
myself in the foot several times doing things like this. Using
`Base.GMP.CdoubleMax` should still be safe since all of those types
can be exactly represented by `Cdouble`.

kalmarek:
That sounds very reasonable, and it's a very good point with potential silent precision loss of floats;
Do you know what is the difference between these two?
```julia
julia> Base.IEEEFloat
Union{Float16, Float32, Float64}

julia> Base.GMP.CdoubleMax
Union{Float16, Float32, Float64}

```
Anyway, either of those is fine.

kalmarek:
I'm happy with this, shall we merge?

saschatimme:
Looks good :)

saschatimme:
Also feel free to merge things without me looking at it in the future. Maybe @Joel-Dahne can get commit rights? (If this is not already the case)

PR #17, opened 2020-05-05
Parse documentation

Joel-Dahne:
So it might be about time that we make use of all the code for parsing Arb-functions. My idea is to automatically generate Julia files from the Arb documentation. Most likely some manual tuning will be required for these files, for example marking some methods as predicates and maybe generalizing some of the argument types. So after automatically generating the files we do the manual tuning and then put it into version control. In case Arb updates the documentation we can generate new files and compare with the ones we already have, it should then be rather easy to do the required updates.

The two commits which are mostly related to this pull request is the on which adds code for parsing the Arb documentation and the one which adds the generated file for Mag. You can read the, short, documentation for the parsing functions. I generated the mag.jl file with

```julia
include("parse/parse.jl")
filename = "../arb/doc/source/mag.rst"; # my place for arb
title, sections = parse_arbdoc(filename);
generate_file(title, sections, filename = "src/mag.jl");
```


saschatimme:
Looks great! I really like this approach. Also since Arb is quite mature, I don’t think we will often need to run this script.

kalmarek:
@Joel-Dahne that's amazing ;) shall I merge or do You want to add something else here?

Joel-Dahne:
We can wait with merging for a while, I have some more updates I want to do. Thought it was a good idea to get something up so that you can give feedback.

One of the things I have thought about is how to distinguish functions that we will never implement (for example random test functions) with functions that we have yet to implement. My idea was to create a list of types that we will not support and somehow mark does functions in the generated code, for example with `##` instead of `#` in front. I have looked a bit at this and so far my list of types that we'll never support, at least to parse automatically, are

```
unsupported_types = Set([
    "FILE *",
    "fmpr_t",
    "fmpr_rnd_t",
    "flint_rand_t",
    "bool_mat_t",
])
```

It's a combination of deprecated types, internal types and types which you would handle differently in Julia. There are probably more to include, for example `acb_calc_func_t`. One benefit with this is that it's clear for us what remains to be done.

Another thing I'll do is to add these generated files for `Arf`, `Arb` and `Acb` as well. Also remove some of the hand-coded versions of them, for example `set!` is written by hand currently.

kalmarek:
another option (that I've been thinking for a while) is to wrap our dictionary `Ctypes` in a `struct` (supporting only `getindex`) and instead of `KeyError` catch `UnsuportedArgumentType` exception. We can then have much finer control over what we throw and under what circumstances.

Joel-Dahne:
That's a good idea! I'll take a look at it.

Joel-Dahne:
Now I have added a struct instead of `Ctypes`. 

My next plan was to add generated files for mag, arf, arb and acb. The bad news are that Arbs documentation contains a lot if minor errors, as an example for the function `void acb_polygamma(acb_t z, const acb_t s, const acb_t z, slong prec)` the argument `z` appears twice (the first `z` is called `res` in the code). I'll try to find these errors and correct them.

I'll also have to remove some of the methods which are already manually written in Arblib, like `set!` and `const_pi!`. That should not be to bad.

Now to the more complicated problem. Some of the functions in Arb has the same name as already existing functions in Julia, for example `cmp`. How should we handle that? We can either explicitly import them into Arblib, in that case we need to be able to detect it, or we can just let them live inside the Arblib namespace. In the latter case we need to make sure than any usage of them inside Arblib are prepended with the correct namespace. What do you think?

kalmarek:
detecting such errors in docs in an automated fashion should be relatively easy (just check if names of CArgs are different), so that's fine;

yes, just remove the hand-written functions whenever in conflict. We need to do some cleaning anyway.

let's first try to separate namespaces. Later we can look into the generated functions signatures and (if we aren't pirating any methods of `Base`) import them accordingly

Joel-Dahne:
Looking more at it I realize that there are a lot of updates that needs to be done to the rest of the code. How about merging this branch now, it implements everything needed to parse the documentation, and then take a new branch for actually incorporating the parsed files into the package?

PR #18, opened 2020-05-07
cleanup

kalmarek:
- [x] remove old macros
- [x] handwritten `set!`
- [x] handwritten predicates

other stuff?


kalmarek:
test will fail (on purpose)

kalmarek:
all `set!` methods in `src/constructors.jl` (lines `1:60`) should eventually go away, other than this is ready to go

Joel-Dahne:
Great! We can merge this and I can start adding the files parsed from the documentation. I'm not experienced with CodeCov, does it give `n/a` because it's the first time it runs now?

kalmarek:
yes, because it doesn't have baseline to compare

PR #20, opened 2020-05-08
set! does not return the result

kalmarek:
this is just to keep compatibility with the new `set!`

after all I think this was a good decision to keep the signatures so close to Arb ;)

### DIFF
--- a/examples/sin_naive.jl
+++ b/examples/sin_naive.jl
@@ -7,7 +7,7 @@ function sin_naive!(res, x)
     # @show tol
     k = 0
     while true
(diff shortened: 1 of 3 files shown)


Joel-Dahne:
I'm currently working on a branch for incorporating the methods parsed from the documentation. I have not gotten to the constructors yet, but I think it's easier to fix it there. I can create a pull request for my current progress of the branch so you can see what I'm doing.

kalmarek:
In your branch you can just delete those 1:60 lines from `constructors.jl`. This fix is only for consistency with the future

Issue #23, opened 2020-05-12
Reconsider finalizer attachment


saschatimme:
Currently `Arb`, `Arf`, etc.  look all like this

```julia
# c memory layout
struct arf_struct
    exponent::UInt      # fmpz
    size::UInt          # mp_size_t
    mantissa1::UInt     # mantissa_struct of length 128
    mantissa2::UInt
end

# mutable struct containing precision
mutable struct Arf <: Real
    arf::arf_struct
    prec::Int
   
   function Arf(si::Int64; prec::Integer=DEFAULT_PRECISION[])
        res = new(arf_struct(0,0,0,0), prec)
        init_set!(res, si)
        finalizer(clear!, res)
        return res
    end
end
```

In particular, the `Arf` struct is mutable since we need to attach a finalizer somewhere. We ended up with this design since this was the only design which only results in one allocation in Julia <= 1.4. Starting from Julia 1.5 however, it is possible to also have

```julia
# c memory layout
mutable struct arf_struct
    exponent::UInt      # fmpz
    size::UInt          # mp_size_t
    mantissa1::UInt     # mantissa_struct of length 128
    mantissa2::UInt

   function Arf(...)
        ... attach finalizer etc.
    end
end

# struct containing precision
struct Arf <: Real
    arf::arf_struct
    prec::Int
end
```
result in only one allocation. 
I think this design question is deeply tied together with how we want to handle indexing into Arb-owned matrices. Using the second approach, we can return an `Arf` without any extra allocations with the caveat that the memory lifetime is controlled by the matrix (but this is the same in C). Our current approach on the other hand would always trigger an allocation (either due to making a copy and attaching a finalizer, or constructing a new mutable struct).

Any thoughts?

kalmarek:
If I remember correctly already 1.4 could elide the additional allocation (1.3 couldn't)

----
well, that's something of a challenge. Julia users will assume that
* `A[i,j]` does not allocate
* `a = A[i,j]` makes `a` outlive `A`.

It seems that without those assumptions using generic code with Arb matrices will be very fragile ?
Consider 
```julia
get_first(m::ArbMatrix) = m[1,1]

a,b = GC.@protect m
    a = get_first(m) # no allocation
    b = get_first(m) + 0  # creates new Arb
    a,b
end
```
here `a` holds garbage (`m` can be `GC`ed anytime after the block), but `b` is independent of `m`.

[...]

PR #25, opened 2020-05-30
Arbcall structs

Joel-Dahne:
I have changed the methods generated by `arbcall` to not only accept the Julia wrapper of the Arb-type but also the corresponding struct as well as pointers to such structs. So for example an argument which would previously take an `Arb` now takes `Union{Arb, arb_struct, Ptr{arb_struct}}`. If the argument is a struct or a pointer the default precision is the current global precision.

The main reason for this is to make it easier to implement non-allocating and inplace operations on vectors and matrices, but also on the radius and midpoint, in the future.

Some more comments:

The last commit fixes a performance issue with `cprefix` that I noticed when comparing the performance after the two commits before that. I believe the problem was that the symbol had to be parsed every time. I tried to figure out how to precompute it in the macro but was not able to get anywhere... So I moved it out and defined it manually instead.

There are not many tests for the new code, in particular for calling the methods with pointers. This will probably be added in a later stage when we add support for `ArbVectors` which will naturally have to use a lot of the pointer methods.

### DIFF
--- a/src/arb_types.jl
+++ b/src/arb_types.jl
@@ -1,16 +1,16 @@
-struct mag_struct
+mutable struct mag_struct
     exponent::UInt       # fmpz
     mantissa::UInt       # mp_limb_t
 end
 
-struct arf_struct
+mutable struct arf_struct
     exponent::UInt      # fmpz
     size::UInt          # mp_size_t
     mantissa1::UInt     # mantissa_struct of length 128
     mantissa2::UInt
 end
 
-struct arb_struct
+mutable struct arb_struct
                         # ┌ arf_struct (midpoint)
     exponent::UInt      # │ fmpz
     size::UInt          # │ mp_size_t
@@ -23,7 +23,7 @@ struct arb_struct
                         # └
 end
 
-struct acb_struct
+mutable struct acb_struct
                           # ┌ arb_struct (real)
     exponent_r::UInt      # │ fmpz
     size_r::UInt          # │ mp_size_t
--- a/src/arb_types.jl
+++ b/src/arb_types.jl
@@ -1,13 +1,41 @@
-mutable struct mag_struct
-    exponent::UInt       # fmpz
-    mantissa::UInt       # mp_limb_t
-end
-
 mutable struct arf_struct
     exponent::UInt      # fmpz
     size::UInt          # mp_size_t
     mantissa1::UInt     # mantissa_struct of length 128
     mantissa2::UInt
+
+    function arf_struct()
+        res = new()
+        init!(res)
+        finalizer(clear!, res)
+        return res
+    end
+
+    function arf_struct(x::Union{UInt, Int})
+        res = new()
+        init_set!(res, x)
+        finalizer(clear!, res)
+        return res
+    end
+end
+
+mutable struct mag_struct
+    exponent::UInt       # fmpz
+    mantissa::UInt       # mp_limb_t
+
+    function mag_struct()
+        res = new()
+        init!(res)
+        finalizer(clear!, res)
+        return res
+    end
+
+    function mag_struct(x::Union{mag_struct, arf_struct})
+        res = new()
+        init_set!(res, x)
+        finalizer(clear!, res)
+        return res
+    end
 end
 
 mutable struct arb_struct
@@ -21,6 +49,13 @@ mutable struct arb_struct
     exponent_mag::UInt  # │ fmpz
     mantissa_mag::UInt  # │ mp_limb_t
                         # └
+
+    function arb_struct()
+        res = new()
+        init!(res)
+        finalizer(clear!, res)
+        return res
+    end
 end
 
 mutable struct acb_struct
@@ -40,4 +75,10 @@ mutable struct acb_struct
     exp_mag_i::Int        # │ fmpz
     mantissa_mag_i::UInt  # │ mp_limb_t
                           # └
+    function acb_struct()
+        res = new()
+        init!(res)
+        finalizer(clear!, res)
+        return res
+    end
 end
--- a/src/arb_types.jl
+++ b/src/arb_types.jl
@@ -41,15 +41,3 @@ mutable struct acb_struct
     mantissa_mag_i::UInt  # │ mp_limb_t
                           # └
 end
-
-for prefix in (:arf, :arb, :acb, :mag)
-    arbstruct = Symbol(prefix, :_struct)
-    arb_init = Symbol(prefix, :_init)
-    arb_clear = Symbol(prefix, :_clear)
-    @eval begin
-        init!(t::$arbstruct) =
-            ccall(@libarb($arb_init), Cvoid, (Ref{$arbstruct},), t)
-        clear!(t::$arbstruct) =
-            ccall(@libarb($arb_clear), Cvoid, (Ref{$arbstruct},), t)
-    end
-end
--- a/src/types.jl
+++ b/src/types.jl
@@ -73,11 +73,6 @@ mutable struct Acb <: Number
     end
 end
 
-cprefix(::Type{Mag}) = :mag
-cprefix(::Type{Arf}) = :arf
-cprefix(::Type{Arb}) = :arb
-cprefix(::Type{Acb}) = :acb
-
 for (T, prefix) in ((Mag, :mag), (Arf, :arf), (Arb, :arb), (Acb, :acb))
     arbstruct = Symbol(prefix, :_struct)
     spref = "$prefix"
@@ -85,6 +80,7 @@ for (T, prefix) in ((Mag, :mag), (Arf, :arf), (Arb, :arb), (Acb, :acb))
         cstructtype(::Type{$T}) = $arbstruct
     end
     @eval begin
+        cprefix(::Type{$T}) = $(QuoteNode(Symbol(prefix)))
         cstruct(x::$T) = getfield(x, cprefix($T))
         Base.convert(::Type{$(cstructtype(T))}, x::$T) = cstruct(x)
     end
--- a/src/types.jl
+++ b/src/types.jl
@@ -3,23 +3,12 @@ mutable struct Arf <: Real
     prec::Int
 
     function Arf(;prec::Integer=DEFAULT_PRECISION[])
-        res = new(arf_struct(0,0,0,0), prec)
-        init!(res)
-        finalizer(clear!, res)
+        res = new(arf_struct(), prec)
         return res
     end
 
-    function Arf(ui::UInt64; prec::Integer=DEFAULT_PRECISION[])
-        res = new(arf_struct(0,0,0,0), prec)
-        init_set!(res, ui)
-        finalizer(clear!, res)
-        return res
-    end
-
-    function Arf(si::Int64; prec::Integer=DEFAULT_PRECISION[])
-        res = new(arf_struct(0,0,0,0), prec)
-        init_set!(res, si)
-        finalizer(clear!, res)
+    function Arf(x::Union{UInt, Int}; prec::Integer=DEFAULT_PRECISION[])
+        res = new(arf_struct(x), prec)
         return res
     end
 end
@@ -28,23 +17,12 @@ mutable struct Mag <: Real
     mag::mag_struct
 
     function Mag()
-        res = new(mag_struct(0,0))
-        init!(res)
-        finalizer(clear!, res)
+        res = new(mag_struct())
         return res
     end
 
-    function Mag(m::Mag)
-        res = new(mag_struct(0,0))
-        init_set!(res, m)
-        finalizer(clear!, res)
-        return res
-    end
-
-    function Mag(arf::Arf)
-        res = new(mag_struct(0,0))
-        init_set!(res, arf)
-        finalizer(clear!, res)
+    function Mag(x::Union{Mag, Arf})
+        res = new(mag_struct(cstruct(x)))
         return res
     end
 end
@@ -54,9 +32,7 @@ mutable struct Arb <: Real
     prec::Int
 
     function Arb(;prec::Integer=DEFAULT_PRECISION[])
-        res = new(arb_struct(0,0,0,0, 0,0), prec)
-        init!(res)
-        finalizer(clear!, res)
+        res = new(arb_struct(), prec)
         return res
     end
 end
@@ -66,9 +42,7 @@ mutable struct Acb <: Number
     prec::Int
 
     function Acb(;prec::Integer=DEFAULT_PRECISION[])
-        res = new(acb_struct(0,0,0,0,0,0, 0,0,0,0,0,0), prec)
-        init!(res)
-        finalizer(clear!, res)
+        res = new(acb_struct(), prec)
         return res
     end
 end
--- a/src/types.jl
+++ b/src/types.jl
@@ -73,6 +73,11 @@ mutable struct Acb <: Number
     end
 end
 
+cprefix(::Type{Mag}) = :mag
+cprefix(::Type{Arf}) = :arf
+cprefix(::Type{Arb}) = :arb
+cprefix(::Type{Acb}) = :acb
+
 for (T, prefix) in ((Mag, :mag), (Arf, :arf), (Arb, :arb), (Acb, :acb))
     arbstruct = Symbol(prefix, :_struct)
     spref = "$prefix"
@@ -80,7 +85,6 @@ for (T, prefix) in ((Mag, :mag), (Arf, :arf), (Arb, :arb), (Acb, :acb))
         cstructtype(::Type{$T}) = $arbstruct
     end
     @eval begin
-        cprefix(::Type{$T}) = Symbol($spref)
         cstruct(x::$T) = getfield(x, cprefix($T))
         Base.convert(::Type{$(cstructtype(T))}, x::$T) = cstruct(x)
     end
--- a/src/types.jl
+++ b/src/types.jl
@@ -7,6 +7,19 @@ mutable struct Arf <: Real
         return res
     end
 
+    function Arf(x::arf_struct;
+                 prec::Integer=DEFAULT_PRECISION[],
+                 shallow::Bool = false)
+        if shallow
+            res = new(x, prec)
+        else
+            res = Arf(prec = prec)
+            set!(res, x)
+        end
+
+        return res
+    end
+
     function Arf(x::Union{UInt, Int}; prec::Integer=DEFAULT_PRECISION[])
         res = new(arf_struct(x), prec)
         return res
@@ -18,6 +31,17 @@ mutable struct Mag <: Real
 
     function Mag()
         res = new(mag_struct())
+        return res
+    end
+
+    function Mag(x::mag_struct;
+                 shallow::Bool = false)
+        if shallow
+            res = new(x)
+        else
+            res = new(mag_struct(x))
+        end
+
         return res
     end
 
@@ -35,6 +59,19 @@ mutable struct Arb <: Real
         res = new(arb_struct(), prec)
         return res
     end
+
+    function Arb(x::arb_struct;
+                 prec::Integer=DEFAULT_PRECISION[],
+                 shallow::Bool = false)
+        if shallow
+            res = new(x, prec)
+        else
+            res = Arb(prec = prec)
+            set!(res, x)
+        end
+
+        return res
+    end
 end
 
 mutable struct Acb <: Number
@@ -43,6 +80,19 @@ mutable struct Acb <: Number
 
     function Acb(;prec::Integer=DEFAULT_PRECISION[])
         res = new(acb_struct(), prec)
+        return res
+    end
+
+    function Acb(x::acb_struct;
+                 prec::Integer=DEFAULT_PRECISION[],
+                 shallow::Bool = false)
+        if shallow
+            res = new(x, prec)
+        else
+            res = Acb(prec = prec)
+            set!(res, x)
+        end
+
         return res
     end
 end
@@ -57,9 +107,5 @@ for (T, prefix) in ((Mag, :mag), (Arf, :arf), (Arb, :arb), (Acb, :acb))
         cprefix(::Type{$T}) = $(QuoteNode(Symbol(prefix)))
         cstruct(x::$T) = getfield(x, cprefix($T))
         Base.convert(::Type{$(cstructtype(T))}, x::$T) = cstruct(x)
-    end
-    T == Mag && continue
-    @eval begin
-        Base.precision(x::$T) = x.prec
     end
 end
--- a/src/types.jl
+++ b/src/types.jl
@@ -1,4 +1,4 @@
-mutable struct Arf <: Real
+struct Arf <: Real
     arf::arf_struct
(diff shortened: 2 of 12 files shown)


test/arbcall-test.jl
```diff
@@ -184,16 +198,27 @@ end
     z = Arb(0)
 
     Arblib.@arbcall_str "void arb_add(arb_t z, const arb_t x, const arb_t y, slong prec)"
-    @test typeof(Arblib.add!(z, x, y)) == Nothing
+    @test Arblib.add!(z, x, y) isa Nothing
+    @test Arblib.add!(
```
kalmarek, line 202: I'm not sure this usage of `pointer_from_obj_ref` and calls to `Arblib.add!` correct? I think that julia metadata header in this call is interpreted as fields of `arb_struct`. maybe there weren't any pointers there, so you got lucky ;)  
```julia
julia> x = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> pl = Ptr{Clong}(pointer_from_objref(x))
Ptr{Int64} @0x00007f1f88ba95f0

julia> for i in 1:8
           @show unsafe_load(pl, i)
       end
unsafe_load(pl, i) = 139773414643248
unsafe_load(pl, i) = 3
unsafe_load(pl, i) = 559108
unsafe_load(pl, i) = 3
unsafe_load(pl, i) = 3
unsafe_load(pl, i) = 0
unsafe_load(pl, i) = 0
unsafe_load(pl, i) = 4294967296

julia> for i in 1:3
           @show unsafe_load(pl, sizeof(UInt)+i)
       end
unsafe_load(pl, sizeof(UInt) + i) = 1
unsafe_load(pl, sizeof(UInt) + i) = 2
unsafe_load(pl, sizeof(UInt) + i) = 3

```
whereas
```julia
julia> px = Base.unsafe_convert(Ptr{Clong}, Base.cconvert(Ptr{Clong}, x)) # exactly what `ccall` does
Ptr{Int64} @0x00007f1f88ba9630

julia> for i in 1:3
           @show unsafe_load(px, i)
       end
unsafe_load(px, i) = 1
unsafe_load(px, i) = 2
unsafe_load(px, i) = 3

```

this is why I said in https://github.com/kalmarek/Arblib.jl/issues/23#issuecomment-627566758 that the layout is not c-compatible and was thinking about implementing our own `unsafe_convert`. This wasn't necessary (there), because (?) `arb_struct` is immutable and stored inline:
```julia
julia> a = Arb("0.1")
[0.1000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.95e-78]

julia> a.arb
Arblib.arb_struct(0xfffffffffffffffd, 0x0000000000000008, 0x0000000000000004, 0x0000000002d2fc60, 0xfffffffffffffefe, 0x0000000020000000)

julia> pa = Base.unsafe_convert(Ref{Arb}, Base.cconvert(Ref{Arb}, a))
Ptr{Arb} @0x00007fefaf5fae10

julia> pa_uint = Ptr{UInt}(pa)
Ptr{UInt64} @0x00007fefaf5fae10

julia> for i in 1:6
           @show unsafe_load(pa_uint, i)
       end
unsafe_load(pa_uint, i) = 0xfffffffffffffffd
unsafe_load(pa_uint, i) = 0x0000000000000008
unsafe_load(pa_uint, i) = 0x0000000000000004
unsafe_load(pa_uint, i) = 0x0000000002d2fc60
unsafe_load(pa_uint, i) = 0xfffffffffffffefe
unsafe_load(pa_uint, i) = 0x0000000020000000

julia> unsafe_load(pa_uint, 7) # our precision field
0x0000000000000100

julia> UInt(precision(a))
0x0000000000000100
```

Looking backwards: do we actually check the value of operations on `Arb`s anywhere in our tests?

kalmarek: On the other hand :)
```julia
julia> pa = Base.unsafe_convert(Ref{Arb}, Base.cconvert(Ref{Arb}, a))
Ptr{Arb} @0x00007fefaf5fae10

julia> pointer_from_objref(a)
Ptr{Nothing} @0x00007fefaf5fae10

```
so this particular call returns the correct pointer? to be honest, this part of julia is still not really clear to me ;/ and I'm confused as to where did the header go?

Joel-Dahne: I had a lot of problems with this, it's very unclear to me how it works exactly. First I tried `pointer_from_objref(x)` but that didn't work, then I tried `pointer_from_objref(x.arb)` and that seemed to work so I went with that. 

But that it makes a difference if `arb_struct` being mutable or not makes sense. That explains some of the problems I encountered when preparing this change (I switched the from immutable to mutable now).  Maybe this change is suboptimal then? Since it changes the memory structure of `Arb` we might encounter problems later on?

Most of the tests I have written do not test the values of the `Arb`s but the test with the examples do test it (`test/examples.jl`). The main reason is that when I wrote most of the tests many of the required methods for checking the results was not implemented. I think this is something that will be fixed when we add a higher level interface.

Also doing some more testing it seems like there is a slight performance penalty to coming from this change :(

src/types.jl
```diff
@@ -73,14 +73,18 @@ mutable struct Acb <: Number
...
+cprefix(::Type{Acb}) = :acb
+
 for (T, prefix) in ((Mag, :mag), (Arf, :arf), (Arb, :arb), (Acb, :acb))
     arbstruct = Symbol(prefix, :_struct)
     spref = "$prefix"
     @eval begin
         cstructtype(::Type{$T}) = $arbstruct
     end
     @eval begin
-        cprefix(::Type{$T}) = Symbol($spref)
```
kalmarek: yeah I was lazy and quoting symbols is nasty: `cprefix(::Type{$T}) = $(QuoteNode(Symbol(prefix))))` gives
```julia
julia> @code_lowered cprefix(Arb)
CodeInfo(
1 ─     return :arb
)
```

src/arb_types.jl
```diff
@@ -1,16 +1,16 @@
-struct mag_struct
+mutable struct mag_struct
```
kalmarek, line 1: why do we need them `mutable`?

src/types.jl
Joel-Dahne: This looks nasty indeed, but I'll change to it!

src/arb_types.jl
saschatimme: That's the reason the `pointer_from_objref` works at all. The object needs to be heap allocated. 

```
  pointer_from_objref(x)

  Get the memory address of a Julia object as a Ptr. The existence of the resulting Ptr will not protect the object
  from garbage collection, so you must ensure that the object remains referenced for the whole time that the Ptr will
  be used.

  This function may not be called on immutable objects, since they do not have stable memory addresses.

  See also: unsafe_pointer_to_objref.
```

But this changes the memory layout of our structs since it now contains another pointer and thus a second indirection :/

Joel-Dahne: I think it needs to be to be able to use it in a `ccall`, but I'm not entirely sure. Removing it breaks things at least, so it is needed in some way.

saschatimme: I think it's fine to have these mutable, bu then we should also attach the finalizer here and change `Arb` etc. to be a `struct`.

kalmarek: but why do we need to get a pointer by calling `pointer_from_objref`? `ccall` doesn't use it, it goes through `cconvert` and `unsafe_convert`

Joel-Dahne: We don't need to use `pointer_from_objref`, so far I have only used it in some tests. But the reason that `pointer_from_objref` doesn't work on immutable data is likely the same reason that `ccall` cannot handle immutable objects.

test/arbcall-test.jl
kalmarek: the slowdown is probably due to the fact that `arb_struct` is no longer stored inline in `Arb`:
```julia
julia> a = Arb("0.1")
[0.1000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.95e-78]

julia> cptr_a = Base.unsafe_convert(Ref{Arb}, Base.cconvert(Ref{Arb}, a))
Ptr{Arb} @0x00007f7c110a4370

julia> cptr_as = Base.unsafe_convert(Ref{Arblib.arb_struct}, Base.cconvert(Ref{Arblib.arb_struct}, a.arb))
Ptr{Arblib.arb_struct} @0x00007f7c12293750

julia> unsafe_load(Ptr{UInt}(cptr_a), 1)
0x00007f7c12293750

```

src/arb_types.jl
kalmarek: well, I probably don't understand what do you mean by saying that `ccall` can handle immutable data.
(the following is happening on `master`, i.e. immutable `arb_struct`):
```julia
julia> a = Arb("0.1")
[0.1000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.95e-78]

julia> b = zero(a)
0

julia> rb = Base.cconvert(Ref{Arblib.arb_struct}, b.arb)
Base.RefValue{Arblib.arb_struct}(Arblib.arb_struct(0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000))

julia> ccall(Arblib.@libarb(arb_add), Cvoid, (Ref{Arblib.arb_struct}, Ref{Arblib.arb_struct}, Ref{Arblib.arb_struct}), rb, a.arb, a.arb)

julia> rb[] #rb was modified since it was passed by reference 
Arblib.arb_struct(0xfffffffffffffffe, 0x0000000000000008, 0x0000000000000004, 0x0000000000eaee40, 0xfffffffffffffeff, 0x0000000020000000)

julia> b.arb
Arblib.arb_struct(0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000)

julia> Arblib.add!(b,a,a)

julia> ccall(Arblib.@libarb(arb_equal), Cint, (Ref{Arblib.arb_struct}, Ref{Arblib.arb_struct}), rb[], zero(a).arb)
0

julia> ccall(Arblib.@libarb(arb_equal), Cint, (Ref{Arblib.arb_struct}, Ref{Arblib.arb_struct}), rb[], b.arb)
1 # rb[] and b.arb are equal

```

Joel-Dahne: But in this case the `arb_struct` is inlined in the `Arb` which is mutable. But I don't really know how this works.

saschatimme: I think all this only works, since `b` is heap allocated, so you have a stable memory address for `b.arb`. If however, you would construct an `arb_struct` from scratch, then it would be only be living in the current function stack. And then you cannot pass this to `ccall` without risking a segfault in my understanding.

kalmarek:
just from the philosophical point of view: I do agree with the idea of passing "naked" Arb structs (or pointers to them) to the functions we define. 
But maybe a better idea is to channel everything through `Ptr{arb_struct}` and (as @saschatimme suggested) attach finalizers there and make `Arb` immutable?

This would require overloading `cconvert(::Type{Ref{arb_struct}}, a::Arb) = cconvert(::Type{Ref{arb_struct}}, a.arb)` but that's easy to do :)

eg.
```julia
struct Arb <: Real
    arb::arb_struct
    prec::Int

    Arb(;prec::Integer=DEFAULT_PRECISION[]) = new(arb_struct(), prec)
end

mutable struct arb_struct
                        # ┌ arf_struct (midpoint)
    exponent::UInt      # │ fmpz
    size::UInt          # │ mp_size_t
    mantissa1::UInt     # │ mantissa_struct of length 128
    mantissa2::UInt     # │
                        # └
                        # ┌ mag_struct (radius)
    exponent_mag::UInt  # │ fmpz
    mantissa_mag::UInt  # │ mp_limb_t
                        # └

    function arb_struct()
        a = new()
        init!(a)
        finalizer(clear!, a)
        return a
    end
end


ctype(::Carg{T}) where T <: Union{Arb}  = Ref{arb_struct}
Base.cconvert(::Type{Ref{arb_struct}}, x::Arb) = Base.cconvert(Ref{arb_struct}, cstruct(x))
```

and

```julia
julia> using Arblib, BenchmarkTools

julia> a = Arblib.Arb("0.1")
[0.1000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.95e-78]

julia> b = zero(a)
0

julia> Arblib.add!(b, a,a); b
[0.20000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 3.89e-78]

julia> @benchmark Arblib.add!($b,$a,$a)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     51.657 ns (0.00% GC)
  median time:      53.995 ns (0.00% GC)
  mean time:        57.738 ns (0.00% GC)
  maximum time:     291.661 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     985

```

julia-1.3.1:
```julia
julia> @benchmark Arblib.add!($b,$a,$a)
BenchmarkTools.Trial: 
  memory estimate:  48 bytes
  allocs estimate:  3
  --------------
  minimum time:     250.930 ns (0.00% GC)
  median time:      267.055 ns (0.00% GC)
  mean time:        289.759 ns (0.24% GC)
  maximum time:     2.768 μs (83.74% GC)
  --------------
  samples:          10000
  evals/sample:     301

```

julia-1.5.0-beta1.0:
```julia
julia> @benchmark Arblib.add!($b,$a,$a)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     51.575 ns (0.00% GC)
  median time:      53.408 ns (0.00% GC)
  mean time:        60.592 ns (0.00% GC)
  maximum time:     233.736 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     985
```

saschatimme:
Julia 1.5 😍

kalmarek:
> Julia 1.5 :heart_eyes: 

it's not different from 1.4.2 in this task :stuck_out_tongue:

saschatimme:
So, I think the best strategy would be then this (just for the `arb` type for simplicity):

1) `arb_struct` is a `mutable struct` and has the finalizer
2) `Arb` is `struct` holding `arb_struct` and `precision::Int`
3) Our functions work with `Union{Arb,arb_struct,Ptr{arb_struct}}`.

Only critical part is how to handle the precision if the assigned value is not an `Arb`. Maybe we require it to be always an `Arb`? Since in 1.5 these can guaranteed be created without any allocations from an `arb_struct` this seems reasonable in my opinion.

Thoughts?

kalmarek:
another takes on precision:
* if the first argument is an `Arb`, then it has it's own precision (and we take it as default); in other cases we require precision to be explicitly specified, or
* define `precision(::arb_struct) = 0` and take `max` of precision of args, throwing an error if it is too small (which forces the user to explicitly pass the precision in the case only `arb_structs are passed`

currently I'm not a big fan of auto-precision, except maybe for our constructors, but that's not a breaking point for me either.

Joel-Dahne:
I like your ideas! Regarding the precision I would be fine with having to explicitly give the precision for `arb_struct` and `Ptr{arb_struct}`, I see them as a low level interface and then it's better to be explicit with them. Since the precision at the moment is only determined by the mutated argument we can easily dispatch on that for handling the different versions.

Joel-Dahne:
Some comments from trying to implement this:

Making `Arb` immutable is problematic because you can no longer change the precision. It's quite common with Arb that you iterate a computation, doubling the precision every time until you get good enough results.

With the code
```julia
using Arblib, BenchmarkTools
a, b = Arb(), Arb();
@benchmark Arblib.add!($b, $a, $a)
```
I get the following results with Julia 1.3.1 (1.4.2 seems to be the same)

On master:
```
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     11.563 ns (0.00% GC)
  median time:      11.574 ns (0.00% GC)
  mean time:        11.761 ns (0.00% GC)
  maximum time:     96.975 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999
```
With the discussed changes (not pushed here yet):
```
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     11.564 ns (0.00% GC)
  median time:      11.574 ns (0.00% GC)
  mean time:        11.681 ns (0.00% GC)
  maximum time:     34.066 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999

```
With the above discussed changes but also making `Arb` mutable:
```
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     13.111 ns (0.00% GC)
  median time:      13.129 ns (0.00% GC)
  mean time:        13.438 ns (0.00% GC)
  maximum time:     45.376 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
```
The timings seems to be a bit unstable but making `Arb` mutable does seem to make things slightly slower.

saschatimme:
> Making Arb immutable is problematic because you can no longer change the precision. It's quite common with Arb that you iterate a computation, doubling the precision every time until you get good enough results.

I think that’s not much of a problem since you can just construct a new `Arb` from the old one with different precision. The underlying arb memory is still the same.

saschatimme:
Also the `2.5ns` are in line with the cost of the additional pointer indirection.

Joel-Dahne:
At the moment we do not have any non-copying constructor for `Arb`, so you cannot create a new `Arb` with updated precision without having to allocate first. We can of course implement it, but I'm not sure how we want to do it, we probably want
```julia
x = Arb()
y = Arb(x, prec = 64)
```
to copy `x`.

saschatimme:
Yes, that's true. But I think we could have a `set_precision(x::Arb, prec::Int)` function which calls an (internal) non-copying constructor.

kalmarek:
ah yes, but since we change into more `arb_struct` oriented design we should probably add new constructors: `Arb(as::arb_struct, prec::Integer)` and maybe even `Arb(a::Arb, prec::Integer)` for setting precision.

I'm surprised that your julia 1.3 doesn't allocate though @Joel-Dahne ?
btw. is my hardware so slow? :D how fast is yours macbook @saschatimme? (5x faster is impressive).

Joel-Dahne:
I'm computing 0 + 0, you are computing 0.1 + 0.1, naturally the former is faster. Since in this case we just want to measure call-time I think 0 + 0 makes more sense to use. Why my 1.3 doesn't allocate I have no idea. This is my versioninfo
```
Julia Version 1.3.1
Commit 2d5741174c (2019-12-30 21:36 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-3770K CPU @ 3.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, ivybridge)
```

I'll take a look at adding a `setprecision` and see how it works. But in that case we have to make it clear that `y = setprecision(x, prec)` makes `x` and `y` share the same underlying `arb_struct`.

I think both `Arb(::arb_struct)` and `Arb(::Arb)` makes sense to add, though I think they should both copy the data.

kalmarek:
> I'm computing 0 + 0, you are computing 0.1 + 0.1, naturally the former is faster. Since in this case we just want to measure call-time I think 0 + 0 makes more sense to use. 

ok, didn't notice this; it's ~`13ns` median here, so comparable.

> Why my 1.3 doesn't allocate I have no idea.

Just to check: did you redefined `Arb` and `arb_struct` as in my comment?

Joel-Dahne:
I think the structs are the same, but I do not redefine `cconvert` but just define `Base.convert(::Type{arb_struct}, x::Arb) = x.arb`. Maybe it's coming from that? I pushed the commit with the mutable version of `Arb` so you can compare with that.

Joel-Dahne:
I pushed a commit which I would like some input on. I have added a constructor for `Arb` (and similar for `Arf` and `Acb`) which takes an `arb_struct` and one that takes another `Arb`. They both have a keyword argument `shallow` which determines if it should make a full copy of the struct or just refer to the one given. So `y = Arb(x, shallow = true)` makes inplace operations on `y` also affect `x` whereas `y = Arb(x, shallow = false)` makes it not. To questions, do you think `shallow` is a good name for this? At the moment it defaults to `false`, is this the right choice?

I also made a `setprecision` which essentially just calls this constructor, it also has a `shallow` argument that defaults to `false`. Is this the right choice?

I also added a similar constructor for `Mag` even though it doesn't really do anything...

kalmarek:
@Joel-Dahne `cconvert` defaults to `convert`, so that's fine;

I'm fine with `shallow` keyword, I thought about incorporating julias `copy` into this, but we'd need to create `copy(a::Arb; prec::Integer)` for this and it immediately feels clumsy (to me); `setprecision` sounds much better.

the choices for the defaults for `shallow` seem fine: constructors should actually construct something, `convert` methods try to avoid constructing whenever possible

Joel-Dahne:
Are we ready to merge this then? I'm not completely convinced everything work out well in the end, but I think we need to start using it more before we know that. Now it's at least simple to implement non-allocating operations on the midpoint and radius
```julia
function midref(x::Arb)
    Ptr{arf_struct}(pointer_from_objref(x.arb))
end

function radref(x::Arb)
    Ptr{mag_struct}(pointer_from_objref(x.arb) + 32)
end
```
And then you can compare the radius of an `x::Arb` with a `m::Mag` using `Arblib.cmp(radref(x), m)`.

kalmarek:
I'm ok with merging this; @saschatimme could you have a look?

btw. even on current master you can 
```julia
julia> function unsafe_radref(x::Arb)
           Ptr{Arblib.mag_struct}(Base.unsafe_convert(Ref{Arb}, Base.cconvert(Ref{Arb}, x)) + 32)
       end

julia> a = Arb("0.1")
[0.1000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.95e-78]

julia> a.arb
Arblib.arb_struct(0xfffffffffffffffd, 0x0000000000000008, 0x0000000000000004, 0x00000000021f2be0, 0xfffffffffffffefe, 0x0000000020000000)

julia> ptr_mag = unsafe_radref(a)
Ptr{Arblib.mag_struct} @0x00007feb5e29eff0

julia> unsafe_load(ptr_mag)
Arblib.mag_struct(0xfffffffffffffefe, 0x0000000020000000)

```
even though `mag_struct` is not mutable, `a::Arb` has stable address and that's enough;
(I'm always wary of using `pointer_from_obj_ref`; and the above example should be wrapped with `GC.@preserve` as well)

Joel-Dahne:
Haha you are right, with `Mag` it worked well before as well. What we could not do before is the midpoint, since `Arf` requires a precision that we could not allocate in a non-allocating way.

saschatimme:
I am happy if we make `Arb`, `Avb` etc. immutable, currently they are still mutable but there is no good reason for that.

Joel-Dahne:
Good catch, I forgot I had not changed this!

saschatimme:
Then let's get this merged :)

PR #27, opened 2020-08-05
Add basic vector and matrix support

saschatimme:
This adds support for matrices. Currently, only `Acb` is implemented, but `Arb` is basically copy paste.

I also took the liberty to format the codebase with [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) which makes the diff quite ugly (but the formatting is contained to a single commit). I also added a gh action to make sure the formatting is consistently kept. I can also split this into two PRs if desired.


src/arbcalls/acb_mat.jl
```diff
@@ -0,0 +1,147 @@
...
+arbcall"void acb_mat_det(acb_t det, const acb_mat_t A, slong prec)"
+arbcall"void acb_mat_approx_solve_triu(acb_mat_t X, const acb_mat_t U, const acb_mat_t B, int unit, slong prec)"
+arbcall"void acb_mat_approx_solve_tril(acb_mat_t X, const acb_mat_t L, const acb_mat_t B, int unit, slong prec)"
+arbcall"int acb_mat_approx_lu(slong * P, acb_mat_t LU, const acb_mat_t A, slong prec)"
+arbcall"void acb_mat_approx_solve_lu_precomp(acb_mat_t X, const slong * perm, const acb_mat_t A, const acb_mat_t B, slong prec)"
+arbcall"int acb_mat_approx_solve(acb_mat_t X, const acb_mat_t A, const acb_mat_t B, slong prec)"
+arbcall"int acb_mat_approx_inv(acb_mat_t X, const acb_mat_t A, slong prec)"
+
+### Characteristic polynomial and companion matrix
+# arbcall"void _acb_mat_charpoly(acb_ptr poly, const acb_mat_t mat, slong prec)"
```
saschatimme, line 123: I am not yet fully sure what an `acb_ptr` is and how we should handle this

src/arbcalls/acb_mat.jl
```diff
@@ -0,0 +1,147 @@
...
+arbcall"void _acb_mat_diag_prod(acb_t res, const acb_mat_t mat, slong a, slong b, slong prec)"
+arbcall"void acb_mat_diag_prod(acb_t res, const acb_mat_t mat, slong prec)"
+
+### Component and error operations
+arbcall"void acb_mat_get_mid(acb_mat_t B, const acb_mat_t A)"
+arbcall"void acb_mat_add_error_mag(acb_mat_t mat, const mag_t err)"
+
+### Eigenvalues and eigenvectors
+# arbcall"int acb_mat_approx_eig_qr(acb_ptr E, acb_mat_t L, acb_mat_t R, const acb_mat_t A, const mag_t tol, slong maxiter, slong prec)"
+# arbcall"void acb_mat_eig_global_enclosure(mag_t eps, const acb_mat_t A, acb_srcptr E, const acb_mat_t R, slong prec)"
```
saschatimme, line 141: Similarly, `acb_srcptr` is a question mark for me.

Joel-Dahne:
This is great! In particular I like the addition of JuliaFormatter!

I have also started some work on matrix support but never got around to finish it. I'll try to look through your work today and tomorrow and also compare it to what I have already done, so far a lot of it looks very similar.

src/arbcalls/acb_mat.jl
Joel-Dahne: `acb_ptr` (and `arb_ptr`) are the vector types in Arb. I started some work on supporting them but it got a bit messy and I switched to implementing matrices. For now I think we can ignore them and add them later on if we implement support for them. `acb_srcptr` is equivalent to `const acb_ptr` so from Julias perspective they are the same.

src/arb_types.jl
```diff
@@ -59,26 +59,41 @@ mutable struct arb_struct
...
 end
+
+mutable struct acb_mat_struct
+
+    entries::Ptr{acb_struct}
+    r::Clong
+    c::Clong
+    rows::Ptr{Ptr{acb_struct}}
+
+    function acb_mat_struct(n::Integer, m::Integer)
```
Joel-Dahne, line 93: I would propose to use `r` and `c` for rows and columns to follow the convention in Arb. Same for the `AcbMatrix` constructor (which currently uses `n` and `m` in the opposite way).

src/types.jl
```diff
@@ -97,7 +90,19 @@ struct Acb <: Number
...
+
+struct AcbMatrix <: AbstractMatrix{Acb}
+    acb_mat::acb_mat_struct
+    prec::Int
+
+    function AcbMatrix(m::Integer, n::Integer; prec::Integer = DEFAULT_PRECISION[])
+        res = new(acb_mat_struct(m, n), prec)
+        return res
+    end
+end
```
Joel-Dahne, line 102: We can also add a construct taking an `acb_mat_struct` as is the case for the other types.
```
    function AcbMatrix(A::acb_mat_struct;
                       prec::Integer=DEFAULT_PRECISION[],
                       shallow::Bool = false)
        if shallow
            res = new(A, prec)
        else
            res = AcbMatrix(A.r, A.c, prec = prec)
            set!(res, A)
        end

        return res
    end
```
should work. I can never decide on if I want to call matrices `A` or `M`, Julia seems to favor `A` in general.

src/arb_types.jl
saschatimme: good point, I will make this consistent

src/types.jl
saschatimme: Sure, but I am not sure how much this is needed. The shallow `Acb` operators are basically only there to support non-allocating indexing of matrices, right?

Joel-Dahne: That's true, it might not be very useful. Then we can skip it for now since it's trivial to add later if we realize it's useful.

src/acb_matrix.jl
```diff
@@ -0,0 +1,50 @@
+function AcbMatrix(A::AbstractMatrix{Acb}; prec::Integer = precision(first(A)))
+    M = AcbMatrix(size(A)...; prec = prec)
+    for j = 1:size(A, 2), i = 1:size(A, 1)
+        M[i, j] = A[i, j]
+    end
+    M
+end
+
+Base.size(M::AcbMatrix) = (M.acb_mat.r, M.acb_mat.c)
+Base.precision(M::AcbMatrix) = M.prec
```
Joel-Dahne, line 10: I would move this to `precision.jl`, also add `AcbMatrix` and `acb_mat_struct` to some of the other methods defined there.

src/acb_matrix.jl
```diff
@@ -0,0 +1,50 @@
+function AcbMatrix(A::AbstractMatrix{Acb}; prec::Integer = precision(first(A)))
+    M = AcbMatrix(size(A)...; prec = prec)
+    for j = 1:size(A, 2), i = 1:size(A, 1)
```
kalmarek, line 3: since you've added `@inbouds` elsewhere maybe here as well?

src/arbcalls/acb_mat.jl
kalmarek: I can take my `AcbVector` prototype and update it for use in here. Should I commit here, or to a separate branch @saschatimme ?

saschatimme:
I also added the `ArbMatrix` now.

saschatimme:
For the remaining bindings, we need to handle the arb vector type.

src/arbcalls/acb_mat.jl
saschatimme: However, you prefer.

saschatimme:
I think I addressed the comments in #28 but:

> We would have to update arbsignature so that it handles acb_srcptr correctly. Now it writes them as const acb_ptr, which is correct C but doesn't match exactly. Basically the computation of c_args needs to be updated to take this into account.

I couldn't reconstruct where this is currently happening.

>  I like the idea of keeping the first underscore in the parsed name. In Arb they mark more unsafe functions for which you have to be careful with the arguments. Julia also uses it in a similar fashion for internal methods. So keeping the underscore would make them stand out more. What do you think?

I think it is fine to not use the underscore for all the vector functions since we carry around the size explicitly. Otherwise, all vector functions have underscores which is also a little bit weird.


**More changes:**
* Also added `ArbVector`
* Moved vector functions from `arb` resp. `acb` into separate files.
* The vector functions all have `len` as an argument. Moved this to a `kwarg` similar to how `prec` is handled.

kalmarek:
it's `arbsignature(af::Arbfunction)` that computes `c_args` string incorrectly

saschatimme:
Ah I see, but this is only during initial generation of the arbcall files right?

kalmarek:
yes, it's only for checking correctness; But I think it's worth fixing it

src/arbcall.jl
```diff
@@ -58,72 +55,85 @@ end
...
-jltype(ca::Carg{Base.MPFR.MPFRRoundingMode}) = Union{Base.MPFR.MPFRRoundingMode, RoundingMode}
+jltype(ca::Carg{arb_rnd}) = Union{arb_rnd,RoundingMode}
+jltype(ca::Carg{Base.MPFR.MPFRRoundingMode}) =
+    Union{Base.MPFR.MPFRRoundingMode,RoundingMode}
 jltype(ca::Carg{Cstring}) = AbstractString
 jltype(ca::Carg{Vector{Clong}}) = Vector{<:Integer}
 jltype(ca::Carg{Vector{Culong}}) = Vector{<:Unsigned}
-jltype(::Carg{T}) where {T <: Union{Mag, Arf, Arb, Acb}}  = Union{T, cstructtype(T), Ptr{cstructtype(T)}}
+jltype(ca::Carg{ArbVector}) = Union{ArbVector,cstructtype(ArbVector)}
+jltype(ca::Carg{AcbVector}) = Union{AcbVector,cstructtype(AcbVector)}
```
saschatimme, line 81: This was in Marek's draft 
```julia
 Union{AcbVector,cstructtype(AcbVector),Ptr{acb_struct}}
```
But the `Ptr{acb_struct}` part could lead to ambiguities in dispatch. Also the default  of `len` didn't work.

saschatimme:
Feel free to go ahead and fix it 🙃

kalmarek:
I'm not sure it's good idea to move vec functions to `arbcall/acb_vec.jl`; `arbcall/acb.jl` is generated automatically by `parse.jl` and It'd be better to keep it this way

As for encoding the reason for commenting an `arbcall"..."`: Maybe we could do in `parse.jl`
```julia
#ut arbcal"..." → unsupported type in signature
#sf arbcal"..." → signature parsing failure
#mo arbcal"..." the reason explained here → manual override with reason ginve in the comment
```
?

saschatimme:
> I'm not sure it's good idea to move vec functions to `arbcall/acb_vec.jl`; `arbcall/acb.jl` is generated automatically by `parse.jl` and It'd be better to keep it this way

Good point. I can move them back.

> As for encoding the reason for commenting an `arbcall"..."`: Maybe we could do in `parse.jl`
> 
> ```julia
> #ut arbcal"..." → unsupported type in signature
> #sf arbcal"..." → signature parsing failure
> #mo arbcal"..." the reason explained here → manual override with reason ginve in the comment
> ```
> 
> ?

Sounds good for me. But the manual override is maintained by a blacklist in `parse.jl` ?

kalmarek:
manual overrides are manual ;) but it's a good point how to maintain them in the long run. maybe a file with manual overrides along with `parse.jl` is a good idea

Joel-Dahne:
> Feel free to go ahead and fix it upside_down_face

I can fix it!

Joel-Dahne:
Regarding the parsing of the signatures I'm currently using a patched version of Arb for this since it has a few mistakes in the documentation. I have gotten some of this merged with Arb but not all if it yet. So I can take care of producing the generated files for now!

Joel-Dahne:
I fixed the warning and also reparsed the whole documentation. The diff is pretty ugly, mostly because we switched how we comment out unsupported calls. I have manually commented out some lines with a `#` without a space after. But I have not yet added a comment about why this is required, at least they are easy to identify.

saschatimme:
Are there some open points left? Otherwise I suggest that we merge this and iterate from there on :)

Joel-Dahne:
I'm up for merging and iterating from there! The only thing remaining that I need is support for Polynomials (and Series) and after that I'll start to actually use it, at which point it will get a lot more testing.

saschatimme:
Yeah, I can actually start to use this after this is merged. So there will probably also be plenty of additional things come up soon :)

PR #28, opened 2020-08-08
Acb vec

kalmarek:
Ok, I expect some discussion on this one: main changes

* add `acb_vec_struct` which does not exist on Arb side
* `acb_vec_struct` contains `length`s, `AcbVector` contains `prec`;
* for now the return type is `AcbVector` which is not even correct;
* alias `acb_vec` and `acb_srcvec` on julia side

Minor changes:
* parse underscore methods
* add exceptions for inplace method with `acb_srcvec` → maybe something better is needed
* indexing to `acb_(vec|mat)_struct` return `Ptr{acb_struct}`, so we can use `set!` directly;

My suggestion: we will probably need to hard-comment some methods (like `print`) would `###` be ok for that + explanation at the end of the line?


src/arb_types.jl
```diff
@@ -83,6 +83,17 @@ mutable struct acb_struct
     end
 end
 
+mutable struct acb_vec_struct
+    entries::Ptr{acb_struct}
+    n::Int
+
```
saschatimme, line 89: Should this get the same memory layout as `acb_mat`? There are functions like
```c
int acb_mat_solve(acb_mat_t X, const acb_mat_t A, const acb_mat_t B, slong prec)
```
which expect a matrix whereas in the Julia mindset a vector should also work. If `acb_vec_struct` has the same memory layout then this is all fine.

saschatimme: Or maybe we don't need `acb_vec_struct` in the first place, just a "vector" construct for `acb_mat`?

kalmarek: I don't quite understand this. Could you elaborate more? Do you mean that we should represent it as
```julia
mutable struct acb_vec_struct
    entries::Ptr{acb_struct}
    r::Int # one of these initialized to 1
    c::Int 
    rows::Ptr{Ptr{acb_struct}} # pointer to entries
    ...
end
```
?

saschatimme: Yeah. I am just thinking about the confusion if we have a dedicated vector type, but which you cannot pass to solving a linear system.

kalmarek: ok, I guess it should be then a row-major vector? i.e. (1 column)

saschatimme: I think so

Joel-Dahne: I would say the solution to this is to write a wrapper for automatic conversion from `AcbVector` to an `AcbMatrix` of the correct shape. There are a lot of similar things about expectations in Julia that we have to write wrappers for I believe.

parse/parse.jl
```diff
@@ -58,25 +67,20 @@ function generate_file(title, sections; filename = nothing)
...
                 f = Arblib.Arbfunction(s)
 
-                @assert s == Arblib.arbsignature(f)
+                s == Arblib.arbsignature(f) || @warn(
+                    "Expected signature: $s\n Obtained signature: $(Arblib.arbsignature(f))")
 
                 str *= "arbcall\"" * s * "\"\n"
             catch e
                 if e isa Arblib.UnsupportedArgumentType
                     str *= "##arbcall\"" * s * "\"\n"
```
Joel-Dahne, line 76: I would propose to make the same change as below, i.e. add a space before `#`

src/acb_vector.jl
```diff
@@ -0,0 +1,35 @@
+@inline clear!(acb_vs::acb_vec_struct) = clear!(acb_vs.entries, acb_vs.n)
+Base.size(acb_v::acb_vec_struct) = (acb_v.n,)
+Base.getindex(v::acb_vec_struct, i::Integer) = v.entries + (i - 1) * sizeof(acb_struct)
```
Joel-Dahne, line 3: Should we add boundary checks here? These are internal functions so it might not be as important. If we do add it we should add explicit `@inbounds` to the internal usages of this.

src/acb_vector.jl
```diff
@@ -0,0 +1,35 @@
...
+function Base.setindex!(v::acb_vec_struct, x, i::Integer)
+    set!(v[i], x)
+    return x
+end
+
+Base.size(v::AcbVector) = size(v.acb_vec)
+Base.precision(v::AcbVector) = v.prec
```
Joel-Dahne, line 10: I would propose to move this to `precision.jl` and also add `AcbVector` to the other corresponding functions there.

src/acb_vector.jl
```diff
@@ -0,0 +1,35 @@
...
+Base.cconvert(::Type{Ptr{acb_struct}}, v::AcbVector) = v.acb_vec
+Base.unsafe_convert(::Type{Ptr{acb_struct}}, v::acb_vec_struct) = v.entries
+
+function AcbVector(v::AbstractVector{Acb}, prec::Integer = precision(first(v)))
+    V = AcbVector(length(v); prec = prec)
+    @inbounds for i in eachindex(V)
```
Joel-Dahne, line 16: This will not work if `v` has non-standard indexing. I would propose a `for x in v` with an increasing `i`.

src/arbcall.jl
```diff
@@ -99,11 +103,11 @@ end
...
 function jlfname(
     arbfname,
-    prefixes = ("arf", "arb", "acb", "mag", "mat"),
-    suffixes = ("si", "ui", "d", "arf", "arb");
+    prefixes = ("arf", "arb", "acb", "mag", "mat", "vec"),
+    suffixes = ("si", "ui", "d", "mag", "arf", "arb", "mpfr", "str");
     inplace = false,
 )
-    strs = split(arbfname, "_")
+    strs = filter(!isempty, split(arbfname, "_"))
```
Joel-Dahne, line 110: I like the idea of keeping the first underscore in the parsed name. In Arb they mark more unsafe functions for which you have to be careful with the arguments. Julia also uses it in a similar fashion for internal methods. So keeping the underscore would make them stand out more. What do you think?

Joel-Dahne:
I also agree that we need a way to handle hard-commenting methods which cannot be automatically handled. Commenting them out with `#` and adding a remark about it is probably enough.

saschatimme:
@Joel-Dahne I merge this branch now into `acb_mat` and then address your comments there.

PR #29, opened 2020-08-11
Bump LoadFlint to 0.3

saschatimme:
I noticed that we are currently at LoadFlint `0.1.3` decided that it is a good idea to print always a warning as soon as GMP_jll is loaded, e.g.,

```julia
julia> using HomotopyContinuation, Arblib
┌ Warning: More than one copy of gmp already loaded, using the 1st one
└ @ LoadFlint ~/.julia/packages/LoadFlint/42zpR/src/LoadFlint.jl:40
```

This was changed to a `@debug` message in 0.2, but I thought just let's try 0.3 directly. Let's see whether this works...

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,4 +11,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 BinaryProvider = "0.5.0"
-LoadFlint = "0.1.0"
+LoadFlint = "0.3.0"


saschatimme:
Obviously not so simple...

kalmarek:
There is already FLINT_jll and Arb_jll:

https://github.com/Nemocas/Nemo.jl/commit/8819be6db2814d250695bb22f98752c193efd90b#diff-dafa28e89e17b597d3651af1dca7481d

saschatimme:
Yeah I also realised that our build.jl is quite out of date... #30

saschatimme:
#31

PR #31, opened 2020-08-11
Enh/use jlls

kalmarek:
I'm not sure if `libarb` is exported from `Arb_jll`, but this seems to work? should we add `const libarb = Arb_jll.libarb` just below `using Arb_jll`?


saschatimme:
I guess, if this works, then there is no need to define this additionally :)

Project.toml
```diff
@@ -4,11 +4,16 @@ authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timm
...
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LoadFlint = "472f376f-f1cf-461b-9ac1-d103423be9b7"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 BinaryProvider = "0.5.0"
-LoadFlint = "0.1.0"
+LoadFlint = "0.3.1"
+Arb_jll = "~2.18.1"
+FLINT_jll = "~2.6.2"
```
saschatimme, line 18: Why do we need the FLINT_jll here?

kalmarek: we don't need it

kalmarek:
ok, but now my question is what is the purpose of `LoadFlint.jl`? It just loads `FLINT_jll` and sets the memory allocation functions, but so does `FLINT_jll`. Compare:
https://github.com/JuliaBinaryWrappers/FLINT_jll.jl/blob/22298d2bc6788ae72cb4ac921d2a71c8593b876a/src/wrappers/x86_64-linux-gnu.jl#L54

https://github.com/oscar-system/LoadFlint.jl/blob/6012ee2c7f553b6cd5aff69da66433a1f621acae/src/LoadFlint.jl#L46

?

saschatimme:
I think this is relevant https://github.com/oscar-system/LoadFlint.jl/issues/7

kalmarek:
I can't really read through that thread, can you summarize what is the difference? I just tested with only `FLINT_jll`, all 1.3, 1.4 and 1.5 pass.

Joel-Dahne:
Does it work to load for example Nemo and Arblib at the same time if we use `Flint_jll` instead?

kalmarek:
please tell me what is the real difference between those `__init__`s :rofl: 
https://github.com/JuliaBinaryWrappers/FLINT_jll.jl/blob/22298d2bc6788ae72cb4ac921d2a71c8593b876a/src/wrappers/x86_64-linux-gnu.jl#L29
https://github.com/oscar-system/LoadFlint.jl/blob/6012ee2c7f553b6cd5aff69da66433a1f621acae/src/LoadFlint.jl#L20

to me they're doing the same thing, but LoadFlint also tries to avoid several `libgmp`s loaded. I don't think it's relevant for `julia>=1.3` since there we explicitly `using GMP_jll`.

Also based on the discussion in the thread linked by @saschatimme (see https://github.com/oscar-system/LoadFlint.jl/issues/7#issuecomment-651648360) it seems that `LoadFlint` will use the broadest compat on `FLINT_jll` and it's up to the packages using it to specify the compat versions.
Maybe then we still need to explicitly depend on `FLINT_jll` for particular version?

PR #32, opened 2020-08-12
Add support for ArbPoly

Joel-Dahne:
So far I have just added one commit with
- ArbPoly and arb_poly_struct types
- Support for parsing arb_poly_t
- Parsed methods from arb_poly.h

One of the main things to discuss is what interface to use for `ArbPoly`. There is [Polynomials.jl](https://github.com/JuliaMath/Polynomials.jl) but I think that might only work for polynomials with a parametrized type. We could maybe still take inspiration from it and possibly overload some of their methods for `ArbPoly`. Do you have any other ideas?

Another thing is that I want to add a dedicated `ArbSeries` type, basically an `ArbPoly` with a specified order. Something like
```
struct ArbSeries
    arb_poly::arb_poly_struct
    order::Int
    prec::Int

    ArbPoly(order::Int; prec::Integer = DEFAULT_PRECISION[]) = new(arb_poly_struct(), order, prec)
end
```
It would work identically to `ArbPoly` when it comes to interfacing directly with functions from Arb. The difference would be that we overload different methods for it. For example `*` would for `ArbPoly` use `arb_poly_mul` whereas for `ArbSeries` it would use `arb_poly_mullow`, for `ArbPoly` the length would depend on the highest non-zero term whereas for `ArbSeries` it would be fixed to the given degree. There are a number of differences like that. Here I'm also not sure what interface to use, we could maybe take inspiration from [TaylorSeries.jl](https://github.com/JuliaDiff/TaylorSeries.jl).


kalmarek:
I'm pretty amazed by how fast we can bootstrap a new Arb type with almost all Arb functionality wrapped!

deps/build.jl
```diff
@@ -0,0 +1 @@
+
```
kalmarek, line 1: don't re-add `build.jl` ;)

src/arbcalls/arb_poly.jl
```diff
@@ -0,0 +1,287 @@
...
+### Memory management
+arbcall"void arb_poly_init(arb_poly_t poly)"
+arbcall"void arb_poly_clear(arb_poly_t poly)"
+arbcall"void arb_poly_fit_length(arb_poly_t poly, slong len)"
+arbcall"void _arb_poly_set_length(arb_poly_t poly, slong len)"
+arbcall"void _arb_poly_normalise(arb_poly_t poly)"
+arbcall"slong arb_poly_allocated_bytes(const arb_poly_t x)"
+
+### Basic manipulation
+arbcall"slong arb_poly_length(const arb_poly_t poly)"
```
saschatimme, line 16: This needs an `import Base: length` at the top

src/arbcalls/arb_poly.jl
```diff
@@ -0,0 +1,287 @@
...
+arbcall"void arb_poly_evaluate_acb(acb_t y, const arb_poly_t f, const acb_t x, slong prec)"
+arbcall"void _arb_poly_evaluate2_horner(arb_t y, arb_t z, arb_srcptr f, slong len, const arb_t x, slong prec)"
+arbcall"void arb_poly_evaluate2_horner(arb_t y, arb_t z, const arb_poly_t f, const arb_t x, slong prec)"
+arbcall"void _arb_poly_evaluate2_rectangular(arb_t y, arb_t z, arb_srcptr f, slong len, const arb_t x, slong prec)"
+arbcall"void arb_poly_evaluate2_rectangular(arb_t y, arb_t z, const arb_poly_t f, const arb_t x, slong prec)"
+arbcall"void _arb_poly_evaluate2(arb_t y, arb_t z, arb_srcptr f, slong len, const arb_t x, slong prec)"
+arbcall"void arb_poly_evaluate2(arb_t y, arb_t z, const arb_poly_t f, const arb_t x, slong prec)"
+arbcall"void _arb_poly_evaluate2_acb_horner(acb_t y, acb_t z, arb_srcptr f, slong len, const acb_t x, slong prec)"
+arbcall"void arb_poly_evaluate2_acb_horner(acb_t y, acb_t z, const arb_poly_t f, const acb_t x, slong prec)"
+arbcall"void _arb_poly_evaluate2_acb_rectangular(acb_t y, acb_t z, arb_srcptr f, slong len, const acb_t x, slong prec)"
```
saschatimme, line 144: Here we should probably change the code gen to produce a function which also accepts an `ArbVector` for the two arguments `arb_srcptr f, slong len`

src/arbcalls/arb_poly.jl
Joel-Dahne: There will be a similar function in `acb_poly.jl` as well. I'm still thinking about how to handle this in a good way.

src/arbcalls/arb_poly.jl
Joel-Dahne: My plan here was to later implement methods which are not in-place for which we add more conveniences, like automatically picking up the length from `AcbVector` and such. It would be great if we could change the code-gen to automatically make length a keyword argument which takes the length of `AcbVector` as default value, but I'm worried this would be hard to do. The length parameter is sometimes called `len`, sometimes `flen` or `glen`, so it's hard to automatically identify it.

saschatimme: Makes sense, but note that I handled some code exactly doing this

> It would be great if we could change the code-gen to automatically make length a keyword argument which takes the length of AcbVector as default value

in #27 for the vector types 🙃

Joel-Dahne:
Now I have finally taken the time to finish this pull request. I have added support for `ArbPoly, ArbSeries, AcbPoly` and `AcbSeries`. So far the implemented methods are quite basic and there is a lot of convenience methods to add later on, but I think this is a good start. I have added tests for everything I have added and also cleaned up some of the other tests a little bit. 

Later on I might do some rearranging of code, moving things from `poly.jl` to `series.jl`, but that can wait for now.

Joel-Dahne:
Now I have merged this with the updated `arbcall`. The merge went very smooth, I really like the changes you did to `arbcall`. The code in the pull request is mostly self contained so it should not be to dangerous to merge it. I have added tests for all new functionality as well as some old functionality. Once this is merged I'll look into parsing more of the Arb header files.

One thing I noted is that now that the length of the vectors become key-word arguments we have some clashes. Specifically `void  _arb_vec_add(arb_ptr C, arb_srcptr A, arb_srcptr B, slong n, slong prec)` and `void _arb_poly_add(arb_ptr C, arb_srcptr A, slong lenA, arb_srcptr B, slong lenB, slong prec)` parses to the same method. My solution now was to comment out the latter. If you do want to add them as polynomials you can always do that by using the length of the shorter one, so I don't see it as a problem for now.

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+) where {T<:Union{ArbPoly,ArbSeries,AcbPoly,AcbSeries}}
+    @boundscheck checkbounds(poly, i)
+    res = eltype(T)(prec = precision(poly))
+    get_coeff!(res, poly, i)
+    return res
+end
+
+Base.@propagate_inbounds function Base.setindex!(
+    poly::Union{ArbPoly,ArbSeries},
+    x::Arb,
```
saschatimme, line 34:
```suggestion
    x::ArbLike,
```

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+    i::Integer,
+)
+    @boundscheck checkbounds(poly, i)
+    set_coeff!(poly, i, x)
+    return x
+end
+
+Base.@propagate_inbounds function Base.setindex!(
+    poly::Union{AcbPoly,AcbSeries},
+    x::Acb,
```
saschatimme, line 44:
```suggestion
    x::AcbLike,
```

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+    i::Integer,
+)
+    @boundscheck checkbounds(poly, i)
+    set_coeff!(poly, i, x)
+    return x
+end
+
+# Constructors
+for (TPoly, T) in [(:ArbPoly, :Arb), (:AcbPoly, :Acb)]
```
saschatimme, line 53:
```suggestion
for (TPoly, T) in [(:ArbPoly, :ArbLike), (:AcbPoly, :AcbLike)]
```

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+    @eval function $TPoly(coeff::$T; prec::Integer = precision(coeff))
+        poly = $TPoly(prec = prec)
+        poly[0] = coeff
+        return poly
+    end
+
+    @eval function $TPoly(
+        coeffs::AbstractVector{$T};
```
saschatimme, line 61:
```suggestion
        coeffs::AbstractVector{<:$T};
```

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+    )
+        poly = $TPoly(prec = prec)
+        @inbounds for i = 1:length(coeffs)
+            poly[i-1] = coeffs[i]
+        end
+        return poly
+    end
+end
+
+for (TSeries, T) in [(:ArbSeries, :Arb), (:AcbSeries, :Acb)]
```
saschatimme, line 72: Prob. the same changes as for the poly case above

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+
+Base.@propagate_inbounds function Base.getindex(
+    poly::T,
+    i::Integer,
+) where {T<:Union{ArbPoly,ArbSeries,AcbPoly,AcbSeries}}
+    @boundscheck checkbounds(poly, i)
+    res = eltype(T)(prec = precision(poly))
+    get_coeff!(res, poly, i)
+    return res
+end
```
saschatimme, line 30: I think should also add a `ref(poly:T, i)` method which returns an `ArbRef`/`AcbRef`

src/poly.jl
```diff
@@ -0,0 +1,119 @@
...
+    one!(res)
+    return res
+end
+
+Base.zero(::Type{T}) where {T<:Union{ArbPoly,AcbPoly}} = T()
+function Base.one(::Type{T}) where {T<:Union{ArbPoly,AcbPoly}}
+    res = T()
+    one!(res)
+    return res
+end
```
saschatimme, line 113:
```suggestion
Base.one(::Type{T}) where {T<:Union{ArbPoly,AcbPoly}} = one!(T())
```

We can do these way shorter by now :)

saschatimme (changes requested):
Looks good! Some comments are nitpicky, but I think we should definitely widen the `setindex!` methods and add a `ref` version for `getindex`.

Joel-Dahne:
I'll wait for #62 to be merged with master and then start updating things!

src/poly.jl
Joel-Dahne: This was my initial plan, but it's slightly complicated. Arb allows getting the value of any index `i >= 0`, for `i` higher than any set index it will just give the value 0. This means that while `poly[i]` will be defined for any `i >= 0` it might not correspond to an actual `arb_struct` in memory.

In practice my experience is that you don't change polynomials element-wise very often, or at least it's definitely not where most of the computational cost lies, so it's not such a big problem.

Joel-Dahne:
I merged it with the updates to master and took care of your comments! The only thing I have not taken care of is adding a `ref` method, but as I explained this is not really applicable.

src/poly.jl
saschatimme: Ok that's a fair point :)

saschatimme (approved):

PR #35, opened 2020-08-13
Grab bag

saschatimme:
As the title suggests this is a grab bag of things I stumbled over / needed.
* Adding non-inplace versions for basic arithmetic, exp, log and trig functions 
* Support for square linear system solving
* Constructor from `Rational`
* Fix `zeros(Acb, n)` and `ones(Acb, n)` (previously this would just allocate one element, i.e., all entries shared the same memory)
* Overload `setindex!(a,b)` to call `set!` where `a` is `Union{Mag, Arf, Arb,Acb}` and `b` a number. This allows the syntax
```julia
x = zero(Acb)
y = Acb(2)
x[] = y # identical to set!(x, y)
```
which is consistent with how, e.g. , `Ref`s are updated.

### DIFF
--- a/src/acb_matrix.jl
+++ b/src/acb_matrix.jl
@@ -1,11 +1,11 @@
-function AcbMatrix(v::AbstractVector; prec::Integer = precision(first(v)))
+function AcbMatrix(v::AbstractVector; prec::Integer = _precision(first(v)))
     M = AcbMatrix(length(v), 1; prec = prec)
     @inbounds for (i, vᵢ) in enumerate(v)
         M[i, 1] = vᵢ
     end
     return M
 end
-function AcbMatrix(A::AbstractMatrix; prec::Integer = precision(first(A)))
+function AcbMatrix(A::AbstractMatrix; prec::Integer = _precision(first(A)))
     M = AcbMatrix(size(A)...; prec = prec)
     @inbounds for j = 1:size(A, 2), i = 1:size(A, 1)
         M[i, j] = A[i, j]
--- a/src/acb_matrix.jl
+++ b/src/acb_matrix.jl
@@ -1,3 +1,10 @@
+function AcbMatrix(v::AbstractVector{Acb}; prec::Integer = precision(first(v)))
+    M = AcbMatrix(length(v), 1; prec = prec)
+    @inbounds for (i, vᵢ) in enumerate(v)
+        M[i, 1] = vᵢ
+    end
+    return M
+end
 function AcbMatrix(A::AbstractMatrix{Acb}; prec::Integer = precision(first(A)))
     M = AcbMatrix(size(A)...; prec = prec)
     @inbounds for j = 1:size(A, 2), i = 1:size(A, 1)
@@ -8,6 +15,10 @@ end
 
 Base.size(A::AcbMatrix) = size(A.acb_mat)
 Base.size(A::acb_mat_struct) = (A.r, A.c)
+
+Base.copy(A::AcbMatrix) = copy!(AcbMatrix(size(A)...; prec=precision(A)), A)
+Base.copy!(A::AcbMatrix, B::AcbMatrix) = (set!(A, B); A)
+Base.copyto!(A::AcbMatrix, B::AcbMatrix)  = (set!(A, B); A)
 
 function Base.getindex(A::acb_mat_struct, i::Integer, j::Integer)
     return ccall(
--- a/src/acb_matrix.jl
+++ b/src/acb_matrix.jl
@@ -37,24 +37,14 @@ Base.@propagate_inbounds function Base.getindex(
     shallow::Bool = false,
 )
     @boundscheck checkbounds(A, i, j)
-    return Acb(unsafe_load(A.acb_mat[i, j]); prec = precision(A), shallow = shallow)
+    return AcbRef(A.acb_mat[i, j], precision(A), cstruct(A))
 end
 
-function Base.setindex!(
-    A::acb_mat_struct,
-    x::Union{Acb,acb_struct,Ref{acb_struct}},
-    i::Integer,
-    j::Integer,
-)
+function Base.setindex!(A::acb_mat_struct, x, i::Integer, j::Integer)
     set!(A[i, j], x)
     return x
 end
-Base.@propagate_inbounds function Base.setindex!(
-    A::AcbMatrix,
-    x::Union{Acb,acb_struct,Ref{acb_struct}},
-    i::Integer,
-    j::Integer,
-)
+Base.@propagate_inbounds function Base.setindex!(A::AcbMatrix, x, i::Integer, j::Integer)
     @boundscheck checkbounds(A, i, j)
     A.acb_mat[i, j] = x
     return x
--- a/src/acb_matrix.jl
+++ b/src/acb_matrix.jl
@@ -16,9 +16,9 @@ end
 Base.size(A::AcbMatrix) = size(A.acb_mat)
 Base.size(A::acb_mat_struct) = (A.r, A.c)
 
-Base.copy(A::AcbMatrix) = copy!(AcbMatrix(size(A)...; prec=precision(A)), A)
+Base.copy(A::AcbMatrix) = copy!(AcbMatrix(size(A)...; prec = precision(A)), A)
 Base.copy!(A::AcbMatrix, B::AcbMatrix) = (set!(A, B); A)
-Base.copyto!(A::AcbMatrix, B::AcbMatrix)  = (set!(A, B); A)
+Base.copyto!(A::AcbMatrix, B::AcbMatrix) = (set!(A, B); A)
 
 function Base.getindex(A::acb_mat_struct, i::Integer, j::Integer)
     return ccall(
--- a/src/acb_matrix.jl
+++ b/src/acb_matrix.jl
@@ -1,11 +1,11 @@
-function AcbMatrix(v::AbstractVector{Acb}; prec::Integer = precision(first(v)))
+function AcbMatrix(v::AbstractVector; prec::Integer = precision(first(v)))
     M = AcbMatrix(length(v), 1; prec = prec)
     @inbounds for (i, vᵢ) in enumerate(v)
         M[i, 1] = vᵢ
     end
     return M
 end
-function AcbMatrix(A::AbstractMatrix{Acb}; prec::Integer = precision(first(A)))
+function AcbMatrix(A::AbstractMatrix; prec::Integer = precision(first(A)))
     M = AcbMatrix(size(A)...; prec = prec)
     @inbounds for j = 1:size(A, 2), i = 1:size(A, 1)
         M[i, j] = A[i, j]
--- a/src/acb_matrix.jl
+++ b/src/acb_matrix.jl
@@ -30,12 +30,7 @@ function Base.getindex(A::acb_mat_struct, i::Integer, j::Integer)
         j - 1,
     )
 end
-Base.@propagate_inbounds function Base.getindex(
-    A::AcbMatrix,
-    i::Integer,
-    j::Integer;
-    shallow::Bool = false,
-)
+Base.@propagate_inbounds function Base.getindex(A::AcbMatrix, i::Integer, j::Integer)
     @boundscheck checkbounds(A, i, j)
     return AcbRef(A.acb_mat[i, j], precision(A), cstruct(A))
 end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -1,23 +1,26 @@
 Base.promote_rule(::Type{Arf}, ::Type{<:Union{AbstractFloat,Integer}}) = Arf
 Base.promote_rule(::Type{Arb}, ::Type{<:Union{AbstractFloat,Integer,Arf}}) = Arb
-Base.promote_rule(::Type{Acb}, ::Type{<:Union{AbstractFloat,Integer,Complex,Arf,Arb}}) = Acb
+Base.promote_rule(
+    ::Type{Acb},
+    ::Type{<:Union{AbstractFloat,Integer,Complex,Arf,Arb,AcbRef}},
+) = Acb
 
 for (jf, af) in [(:+, :add!), (:-, :sub!), (:*, :mul!), (:/, :div!)]
     @eval function $(Expr(:., :Base, QuoteNode(jf)))(
         x::T,
         y::T,
-    ) where {T<:Union{Arf,Arb,Acb}}
+    ) where {T<:Union{Arf,Arb,Acb,AcbRef}}
         z = T(prec = max(precision(x), precision(y)))
         $af(z, x, y)
         z
     end
 end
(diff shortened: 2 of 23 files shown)


Joel-Dahne:
We should also maybe add `ArbRef`, similarly to `AcbRef`.

src/arithmetic.jl
```diff
@@ -0,0 +1,59 @@
+Base.promote_rule(::Type{Arf}, ::Type{<:Union{AbstractFloat,Integer}}) = Arf
+Base.promote_rule(::Type{Arb}, ::Type{<:Union{AbstractFloat,Integer,Arf}}) = Arb
+Base.promote_rule(
+    ::Type{Acb},
+    ::Type{<:Union{AbstractFloat,Integer,Complex,Arf,Arb,AcbRef}},
+) = Acb
+
+for (jf, af) in [(:+, :add!), (:-, :sub!), (:*, :mul!), (:/, :div!)]
+    @eval function $(Expr(:., :Base, QuoteNode(jf)))(
```
kalmarek, line 9: I think it's enough to `@eval function Base.$jf(...` here?

saschatimme: Mhm, I remember getting an error message doing it this way, but I just double checked and this indeed works :)

saschatimme:
I added the `ArbRef` implementation and some more minor things.

src/arithmetic.jl
```diff
@@ -0,0 +1,56 @@
...
+    :asec,
+    :sech,
+    :asech,
+]
+    @eval function Base.$f(x::T) where {T<:Union{Arf,Arb,ArbRef,Acb,AcbRef}}
+        z = T(prec = max(precision(x)))
+        $(Symbol(f, :!))(z, x)
+        z
+    end
+end
```
Joel-Dahne, line 56: `Arf` implements surprisingly few functions, basically only addition, subtraction, multiplication, division and square root. In particular it doesn't implement any trigonometric functions. Since they are so limited we should maybe consider treating `Arf` more of an internal type, similar to `Mag`. We could of course convert to `Arb`, preform the operation, and convert back (this is how [ArbFloats.jl](https://github.com/JuliaArbTypes/ArbFloats.jl) does it) but that sort of destroys the point.

src/types.jl
```diff
@@ -7,14 +7,9 @@ struct Arf <: Real
...
-        if shallow
-            res = new(x, prec)
-        else
-            res = Arf(prec = prec)
-            set!(res, x)
-        end
-
+    function Arf(x::arf_struct; prec::Integer = DEFAULT_PRECISION[])
+        res = Arf(prec = prec)
+        set!(res, x)
```
Joel-Dahne, line 12: I see you have removed the `shallow` parameter here and in other places. I guess this is because we are now using `ArbRef` and `AcbRef`? Doesn't the `shallow` parameter still make sense if you want to copy, for example, an `Arb` only increasing the precision of it? I guess this might not be something you do in a tight loop in the same way as iterating through an `ArbVec` or `ArbMatrix`, so it might not be as important to optimise for though.

kalmarek:
In general I really like the concept of `ArbRef`!

src/acb_matrix.jl
```diff
@@ -1,4 +1,11 @@
-function AcbMatrix(A::AbstractMatrix{Acb}; prec::Integer = precision(first(A)))
+function AcbMatrix(v::AbstractVector; prec::Integer = precision(first(v)))
```
Joel-Dahne, line 1: `precision(first(v))` will not always be defined and in general I believe it's only for types from `Arblib.jl` that we want to determine the precision in this way, possibly also `BigFloat`. But in general we should use the global precision.

src/types.jl
```diff
@@ -57,18 +47,36 @@ struct Arb <: Real
...
     end
 end
 
+
+struct ArbRef <: Number
+    arb_ptr::Ptr{arb_struct}
+    prec::Int
+    parent::Union{arb_vec_struct,arb_mat_struct}
+end
+function ArbRef(
```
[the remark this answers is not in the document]
Joel-Dahne: Well in general `prec` is a keyword argument.

Joel-Dahne:
> In general I really like the concept of `ArbRef`!

Same here! It seems to work really well, though I have yet to try it.

Joel-Dahne:
We could maybe merge `arb_matrix.jl` and `acb_matrix.jl` into `matrix.jl` since most of the functionality is identical.

src/constructors.jl
```diff
@@ -68,15 +75,21 @@ for T in (Unsigned, Integer, Base.GMP.CdoubleMax)
         end
     end
 end
+function Acb(x::Arf; prec::Integer = precision(x))
+    res = Acb(prec = prec)
+    # There is not set! with Acb and Arf. So create intermediate Arb :shrug:
+    set!(res, Arb(x, prec = prec))
+    return res
+end
```
Joel-Dahne, line 83: If we want to microoptimize we could construct only an `arb_struct`, though from Julia version 1.5 this might be optimized away anyway.

src/constructors.jl
```diff
@@ -58,6 +58,13 @@ function Arb(str::AbstractString; prec::Integer = DEFAULT_PRECISION[])
     return res
 end
 
+function Arb(x::Rational; prec::Integer = DEFAULT_PRECISION[])
+    num = Arb(numerator(x); prec = prec)
+    denom = Arb(denominator(x); prec = prec)
+    div!(num, num, denom)
+    return num
+end
+
```
Joel-Dahne, line 67: Maybe Arb already optimizes division of integers but if not it might be beneficial to construct and `fmpq` from Flint and then set the value using that. But this is more work than I think is reasonable to do at the moment.

saschatimme: Yes, maybe I should have added a comment. This is definitely better, but I didn't want to figure out how this can be done. So I opted for at least a correct conversion :)

src/types.jl
saschatimme: Yes, I removed this since this primarily used for the vector / matrix indexing. I have some thoughts about the semantics of changing the precision, but I didn't want to drag this in here.

src/types.jl
```diff
@@ -149,3 +174,21 @@ for (T, prefix) in (
...
+cprefix(::Type{ArbRef}) = :arb_struct
+cstructtype(::Type{ArbRef}) = Ptr{arb_struct}
+cstruct(x::ArbRef) = x.arb_ptr
+Base.convert(::Type{Ptr{arb_struct}}, x::ArbRef) = cstruct(x)
+Base.cconvert(::Type{Ref{arb_struct}}, x::ArbRef) = cstruct(x)
+
+cprefix(::Type{AcbRef}) = :acb_struct
+cstructtype(::Type{AcbRef}) = Ptr{acb_struct}
+cstruct(x::AcbRef) = x.acb_ptr
+Base.convert(::Type{Ptr{acb_struct}}, x::AcbRef) = cstruct(x)
```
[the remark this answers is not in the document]
saschatimme: I guess only for ccalls. So then `unsafe_convert` is the correct way to do?

src/acb_matrix.jl
saschatimme: So we should define an internal helper function which falls back to the global precision?

src/arithmetic.jl
saschatimme: `Arf` is really an add one. I will just remove `Arf` from the list, and I think the conversion back and forth should be done explicitly.

src/constructors.jl
saschatimme: I don't think this conversion happens that often at all.

saschatimme:
> We could maybe merge `arb_matrix.jl` and `acb_matrix.jl` into `matrix.jl` since most of the functionality is identical. Same for `arb_vector.jl` and `acb_vector.jl` into `vector.jl`.

Yes, I don't really like the current file structure either. I just wanted to avoid putting everything into this one PR. 😅

Joel-Dahne:
> > We could maybe merge `arb_matrix.jl` and `acb_matrix.jl` into `matrix.jl` since most of the functionality is identical. Same for `arb_vector.jl` and `acb_vector.jl` into `vector.jl`.
> 
> Yes, I don't really like the current file structure either. I just wanted to avoid putting everything into this one PR. sweat_smile

That's a good point! Maybe keep it as it is for now and then we can always clean up later on.

src/acb_matrix.jl
Joel-Dahne: That might be a good solution! Something like 
```
@inline _precision(x::ArbTypes) = precision(x)
@inline _precision(_) = DEFAULT_PRECISION[]
```
might be good enough? But then we should try to use this only when necessary and prefer `precision` otherwise I think.

src/types.jl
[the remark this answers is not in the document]
Joel-Dahne: Well it does make sense in general right? The given definition is the only sensible way to convert a `AcbRef` to a `Ptr{acb_struct}`.

src/types.jl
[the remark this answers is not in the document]
kalmarek: sure, didn't spot the semicolon;)

src/types.jl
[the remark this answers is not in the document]
kalmarek: This is what `essentials.jl` say:
https://github.com/JuliaLang/julia/blob/c6f1f1a616c116e2782995b0e7fe667f7f3baec0/base/essentials.jl#L397
```julia
cconvert(T::Type, x) = convert(T, x) # do the conversion eagerly in most cases
cconvert(::Type{<:Ptr}, x) = x # but defer the conversion to Ptr to unsafe_convert
unsafe_convert(::Type{T}, x::T) where {T} = x # unsafe_convert (like convert) defaults to assuming the convert occurred
unsafe_convert(::Type{T}, x::T) where {T<:Ptr} = x  # to resolve ambiguity with the next method
unsafe_convert(::Type{P}, x::Ptr) where {P<:Ptr} = convert(P, x)
```

saschatimme:
Is there something left to address here?

Joel-Dahne:
> Is there something left to address here?

Let me go through it again now!

src/arithmetic.jl
```diff
@@ -0,0 +1,56 @@
+Base.promote_rule(::Type{Arf}, ::Type{<:Union{AbstractFloat,Integer}}) = Arf
+Base.promote_rule(::Type{Arb}, ::Type{<:Union{AbstractFloat,Integer,Arf,ArbRef}}) = Arb
+Base.promote_rule(
+    ::Type{Acb},
+    ::Type{<:Union{AbstractFloat,Integer,Complex,Arf,Arb,ArbRef,AcbRef}},
+) = Acb
```
Joel-Dahne, line 6: We could also promote `Rational` and do we have to specify which subtypes of `Complex` we want to promote?

src/arithmetic.jl
```diff
@@ -0,0 +1,56 @@
...
+function Base.:(-)(x::T) where {T<:Union{Arf,Arb,ArbRef,Acb,AcbRef}}
+    z = T(prec = precision(x))
+    neg!(z, x)
+    z
+end
+function Base.:(^)(x::T, k::Integer) where {T<:Union{Arf,Arb,ArbRef,Acb,AcbRef}}
+    z = T(prec = precision(x))
+    pow!(z, x, convert(UInt, k))
+    z
+end
```
Joel-Dahne, line 24: `Arf` doesn't support `pow!`

src/linear_algebra.jl
```diff
@@ -0,0 +1,56 @@
+for (jf, af) in [(:+, :add!), (:-, :sub!), (:*, :mul!)]
+    @eval function Base.$jf(A::T, B::T) where {T<:Union{ArbMatrix,AcbMatrix}}
+        C = T(size(A, 1), size(B, 2); prec = max(precision(A), precision(B)))
+        $af(C, A, B)
+        C
+    end
+end
```
Joel-Dahne, line 7: We should check that the dimensions agree here and throw an error otherwise. This would be done differently for addition and multiplication.

Joel-Dahne: Similar things should be done for some of the other methods.

tst.jl
```diff
@@ -0,0 +1,8 @@
+using Pkg
+Pkg.activate(@__DIR__)
+using Arblib, LinearAlgebra, BenchmarkTools
+
+
+v = AcbVector(3)
+
+Arblib.mul!(c, a, b)
```
Joel-Dahne, line 8: Is this file supposed to be here?

Joel-Dahne (changes requested):
I have some more comments, but in general I think we can merge this!

tst.jl
saschatimme: 😅 no

src/arithmetic.jl
saschatimme: I aded the promotion from Rational, and restricted `Complex` to `Complex{<:Union{AbstractFloat,Integer,Rational}}`

src/arithmetic.jl
saschatimme: Thanks, I shouldn't have defined anything for `Arf` 🙃

src/linear_algebra.jl
saschatimme: I added bounds check for all the linear algebra routines

saschatimme:
I rebased onto master, and rewrote the history to hide my mess-yp with the `tst.jl` file.

src/linear_algebra.jl
```diff
@@ -0,0 +1,91 @@
+for (jf, af) in [(:+, :add!), (:-, :sub!)]
+    @eval function Base.$jf(A::T, B::T) where {T<:Union{ArbMatrix,AcbMatrix}}
+        @boundscheck (
+            size(A) == size(B) ||
+            throw(DimensionMismatch("Matrix sizes are not compatible."))
+        )
```
Joel-Dahne, line 6: I don't think it's common to use `@boundscheck` here, from what I can tell `Base` doesn't do it.

Joel-Dahne: Similarly for the other cases where this is done.

Joel-Dahne: Otherwise the check looks good!

saschatimme:
I forgot to implement the `_precision` suggestion from you, just added it now.

src/linear_algebra.jl
saschatimme: Well, this just means that an @inline will elide this check. Not that it matters much at all.

saschatimme: I also think this is used quite a lot in Base, e.g., https://github.com/JuliaLang/julia/blob/cebd4fa959b6f9835a3abe70e705136b5e914e4a/stdlib/LinearAlgebra/src/bitarray.jl#L98

Joel-Dahne: Okay Base seems to vary a bit, lets keep it then!

Joel-Dahne (approved):

saschatimme:
Awesome 👍  I try to keep the next pull requests smaller :)

PR #36, opened 2020-08-13
Delete build.jl

saschatimme:


saschatimme:
Seems that I accidentally recreated this by merging master into #31

Issue #16, opened 2020-04-19
What to return from arbcall?


kalmarek:
this is just to have a space for discussion on the topic rised in [here](https://github.com/kalmarek/Arblib.jl/pull/15#issuecomment-616193596):
> @Joel-Dahne wrote:
I think they case of `Bool` versus `Int32` is interesting as well. While many of the int types in arb represents booleans, some of them are flags with more possible values. It would fit much better in Julia to return a `Bool` instead of an `Int32` when possible, but then we need some way of determining when this can be done. When it occurs as an argument we can use `Union{Bool, Int32}` or maybe even `Integer`.

Another solution is to specialize `@arbcall macro on classes of functions, (one of them: predicates), e.g. 
```julia
@arbcall "arb_add(...)" # normal function
@arbcall Predicate "int arb_is_zero(const arb_t x)" iszero
```
which should call
```julia
macro arbcall(ftype::Expr, arbsignature_str::String,
    jl_fname=jlfname(Arbfunction(arbsignature_str)))
    af = Arbfunction(arbsignature_str)
    local T = eval(ftype)
    return jlcode(T, af, jl_fname)
end

function jlcode(::Type{Predicate}, af::Arbfunction, jl_fname=jlfname(af))
    returnT = returntype(af)
    cargs = arguments(af)
    args, kwargs = jlargs(af)

    call = :(
        function $jl_fname($(args...); $(kwargs...))
              res = ccall(Arblib.@libarb($(arbfname(af))),
              $returnT,
              $(Expr(:tuple, ctype.(cargs)...)),
              $(Symbol.(name.(cargs))...))
              return !iszero(res)
        end
    )
end
```
(I don't like this `eval`, but i'm not sure what to do about it)

kalmarek:
one potential idea is to have a look at jump macros to see how do they handle such constructs:
```julia
m = Model()
@variable m x[1:5]
```

Joel-Dahne:
You are referring to the Julia JuMP package? I'm not entirely sure what you mean, what is it that we should see how they handle?

kalmarek:
yes, the JuMP package; 
they have a set of macros: `@variable`, `@constraint`, `@objective` that take as first argument an object (a model) and dispatch somehow on its type (e.g. you can have `PolyModel()` for polynomial opimization, etc.). 
But anyway, maybe it's the time to test our macros in practice and start producing actual working functions ;)

Joel-Dahne:
[...]

PR #37, opened 2020-08-13
Return first argument in arbcall if return type is void and inplace func

saschatimme:
Addresses my comment in https://github.com/kalmarek/Arblib.jl/issues/16#issuecomment-673342053.
I think this fits more the idiomatic Julia code, and should be a no-op anyways.

### DIFF
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -260,11 +260,18 @@ function jlcode(af::Arbfunction, jl_fname = jlfname(af))
 
     return :(
         function $jl_fname($(args...); $(kwargs...))
-            ccall(
+            __ret = ccall(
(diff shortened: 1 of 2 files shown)


Joel-Dahne:
This should be safe to do and as long as it's clear when it happens I think it's fine to do. There are several cases this doesn't apply to, for example all functions on `Arf` since does return information about rounding. So this will lead to some asymmetries, but that might be okay.

saschatimme:
> There are several cases this doesn't apply to, for example all functions on Arf since does return information about rounding. So this will lead to some asymmetries, but that might be okay.

I think these asymmetries are okay. If you want you can ignore the return value with out no cost.

kalmarek (approved):

PR #42, opened 2020-08-20
use FLINT_jll directly

kalmarek:


saschatimme:
Can you tell me why we need to depend directly on `Flint_jll` if this is already used in the `Arb_jll`?

kalmarek:
Ha! great question! ;D maybe we don't and just import its `libflint` ;)

saschatimme (approved):

Joel-Dahne:
No better feeling than removing dependencies!

PR #44, opened 2020-08-21
convert(::Type{<:ArbRef/AcbRef})

saschatimme:


kalmarek:
tbh, I'm slightly against this non-standard conversion, but let's have a general look at julia conventions we're breaking:
```julia
M = ArbMatrix(rand(2,3)) # ArbMatrix
M[1,2] # ArbRef
x = M[1,2]
x = 3 # modifies M
y = Arb(M[2,1]) # honest Arb
z = y
z[] = 3 # modifies y
```
The second one is harmless (and cool:) since nobody unaware of Arblib will use this syntax; The first one will lead to a number of problems; 
How about making `M[1,2]` return a honest `Arb`, but introducing `M[]` as a `Ref{ArbMatrix}` and `M[][1,2]` resulting in `ArbRef`?

kalmarek:
tbh, I'm slightly against this non-standard conversion, but let's have a general look at julia conventions we're breaking:
```julia
M = ArbMatrix(rand(2,3)) # ArbMatrix
x = M[1,2] # ArbRef
add!(x, 2) # modifies M
y = Arb(M[2,1]) # honest Arb
add!(y,2) # modifies y, but not M
y[] = 3 # modifies y, but not M
```
The second one is harmless (and `y[] = 2` is cool:) since nobody unaware of Arblib will use this syntax; The first one will lead to a number of problems; How about making `M[1,2]` return a honest `Arb`, but introducing
```julia
Base.getindex(M::ArbMatrix) = Ref(M)
Base.getindex(M::Ref{ArbMatrix}, i, j) = ArbRef(M[].arb_mat, i, j)`
``` 
thus `eltype(M)` can stay `Arb`, but we also have access to allocation-less `M[][1,2]` resulting in `ArbRef`?

Joel-Dahne:
Isn't that how it works for normal arrays as well though?
```
M = Arb.(rand(2, 3))
x = M[1,2]
Arblib.add!(x, x, 2)
```
will also modify `x`. Numerical code in Julia is in general not mutating numbers but always overwriting them. For example `BigFloat` has not in-place methods supported. So this behaviour is what you would expect if you choose to use in-place methods. Using `x = x + 2` would not mutate `M` since in that case we are creating a new variable (how we handle `x += 2` is a story for another day).

kalmarek:
will also modify `M` ;) yes, and that makes me a little bit uneasy, but maybe it shouldn't ;)

In any case the `M[][1,2]` syntax is a solution to the `eltype` problem which also keeps the flexibility; but it's just an idea.

btw. Do we have a choice on how we handle `x += 2`? from what I understand this is lowered to `x = x + 2`?

kalmarek:
ok, the `Ref{ArbMatrix}` is not great since we loose the type information (e.g. it's not longer a vector);

The other solution is to create `RefVector{Arb}`/`ArbRefVector` (more types, yay!) returning `ArbRef`s for those who know how to drive and a safe bus seat of `ArbVector` for generic algorithms?

Joel-Dahne:
Isn't the current implementation of `ArbVector` as safe as any normal `Vector` at the moment? It would of course look cleaner if it returned `Arb` instead of `ArbRef`.

kalmarek:
current `ArbVector` isn't as safe as `Vector`, since the elements carry a reference to `arb_mat` and their values are still linked to it. It's true that currently using standard arithmetic it's easy to break this link but this (contrived) example shows that this link may leak information and completely change the result:
```julia
julia> function normalize!(vec::AbstractVector, i::Integer)
           n = vec[i]
           for j in eachindex(vec)
              vec[j] = vec[j]/n
           end
           return n
       end
normalize! (generic function with 1 method)

julia> v = Float64[1:5...]; v_arb = ArbVector(v);

julia> n = normalize!(v, 3); n_arb = Arb(normalize!(v_arb, 3)); (n, n_arb)
(3.0, 1.0000000000000000000000000000000000000000000000000000000000000000000000000000)

julia> v .≈ Arb.(v_arb)
5-element BitArray{1}:
 1
 1
 1
 0
 0

julia> v
5-element Array{Float64,1}:
 0.3333333333333333
 0.6666666666666666
 1.0
 1.3333333333333333
 1.6666666666666667

julia> v_arb
5-element ArbVector:
 [0.33333333333333333333333333333333333333333333333333333333333333333333333333333 +/- 4.78e-78]
 [0.66666666666666666666666666666666666666666666666666666666666666666666666666666 +/- 9.55e-78]
                 4.0000000000000000000000000000000000000000000000000000000000000000000000000000
                 5.0000000000000000000000000000000000000000000000000000000000000000000000000000

```

saschatimme:
I think @kalmarek's example also shows that we currently cannot expect an `ArbVector` to work in a generic algorithm. Thus, I agree that we should split this into a safe default (`ArbVector`) which returns an honest `Arb` (slower but reliable) and more unsafe (and closer to the C-api)  version (`ArbRefVector`) which returns an `ArbRef`. Converting between these two is easy and we can widen the codegen to cover both.
If you @Joel-Dahne @kalmarek agree with this design, then I would go ahead and provide a pull request.

kalmarek:
what is even worse is that
```julia
julia> function normalize2!(vec::AbstractVector, i::Integer)
           n = vec[i]
           vec ./= n
           return n
       end
```
produces incorrect but different results ;)

on the other hand
```julia
julia> function normalize2!(vec::AbstractVector, i::Integer)
           n = vec[i] + Arb(1) # allocates new Arb
           vec ./= n
           return n
       end
end
```
should be correct, since the link between `n` and `vec` is broken. This is very hard to explain to users and is easy to see how it may lead to many subtle bugs

Joel-Dahne:
Okay this does look very dangerous and should definitely be avoided! Making two different interfaces is probably the better choice then. Could it maybe be enough to just implement `ref(v::ArbVector, i::Integer)` to get an `ArbRef` instead? It makes the indexing more annoying but if you really want to perform in place operations then it might be acceptable anyway?

PR #49, opened 2020-09-01
Add also set!(z::Union{AcbRef,Ptr{acb_struct}},x::Complex{<:Base.GMP.CdoubleMax})

saschatimme:
Follow-up to #47 which only handled `Acb` as the assigned. This widens this to also allow `AcbRef` and `Ptr{acb_struct}`.


Joel-Dahne (commented):
You can add `acb_struct` there as well!

saschatimme:
@Joel-Dahne I also added `acb_struct` :)

Joel-Dahne (approved):

PR #51, opened 2020-09-02
Add realref, imagref, real, imag of Acb

saschatimme:
Extends the parent field of `ArbRef` to also allow `Acb`. The `realref` 
and `imagref` methods are recommended in the Arb manual as a fallback if 
a method is not implemented for `Acb`.

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -68,3 +68,15 @@ for f in [
         z
     end
 end
+
+function realref(z::AcbLike; prec = _precision(z))
+    real_ptr = ccall(@libarb(acb_real_ptr), Ptr{arb_struct}, (Ref{acb_struct},), z)
+    ArbRef(real_ptr, prec, cstruct(z))
+end
+function imagref(z::AcbLike; prec = _precision(z))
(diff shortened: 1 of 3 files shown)


kalmarek (changes requested):
note that since we return the first argument of modifying calls we can write these methods as
```julia
Base.real(z::Union{...}) = Arblib.get_real!(Arb(prec = _precision(z)), z)
Base.imag(z::Union{...}) = Arblib.get_imag!(Arb(prec = _precision(z)), z)
```
as I did yesterday ;) Can You add here
```julia
Base.conj(z::Union{...}) = conj!(Acb(prec = _precision(z)), z)
```
which is needed for matrix `'/`adjoint` ;)

btw. does `JuliaFormatter` require `return` keyword? when I hit `Ctrl+Shift+I` in VSCode I get those added everywhere...

saschatimme:
@kalmarek I shortened the definitions and added the `conj` definition.

> btw. does JuliaFormatter require return keyword? when I hit Ctrl+Shift+I in VSCode I get those added everywhere...
This

kalmarek (approved):

PR #52, opened 2020-09-02
Add ArfRef

saschatimme:
This adds a type `ArbRef` similar to `ArbRef` and `AcbRef`. The main motivation is to obtain the midpoint of an `Arb` as a `Float64` without having to allocate any temporaries.
Together with #51 this also allows to convert `Acb` to `ComplexF64`.


Joel-Dahne:
This is a nice addition! I guess it will change a bit now after merging #57, mostly making it simpler. There are some lines which do not have tests, would be good to add that as well.

PR #53, opened 2020-09-02
Convert Mag, Arf to Float64

saschatimme:
This allows to convert a `Mag` or `Arf` to a `Float64` using the syntax 
`m[]` which is quite neat. The other types don't have a simple `get` 
function.

### DIFF
--- a/src/Arblib.jl
+++ b/src/Arblib.jl
@@ -14,8 +14,8 @@ macro libflint(function_name)
 end
 
 include("arb_types.jl")
-include("types.jl")
 include("rounding.jl")
+include("types.jl")
 include("arbcall.jl")
 
 include("precision.jl")
--- a/src/types.jl
+++ b/src/types.jl
@@ -187,3 +187,17 @@ function Base.setindex!(x::Union{Mag,Arf,Arb,ArbRef,Acb,AcbRef}, z::Number)
     set!(x, z)
     x
 end
+
+Base.Float64(x::Mag) = get(x)
+function Base.Float64(
+    x::Union{Arf,Ptr{arf_struct}};
+    rnd::Union{arb_rnd,RoundingMode} = RoundNearest,
+)
+    ccall(@libarb(arf_get_d), Cdouble, (Ref{arf_struct}, arb_rnd), x, rnd)
+end
+function Base.Int(
+    x::Union{Arf,Ptr{arf_struct}};
+    rnd::Union{arb_rnd,RoundingMode} = RoundNearest,
+)
+    ccall(@libarb(arf_get_si), Clong, (Ref{arf_struct}, arb_rnd), x, rnd)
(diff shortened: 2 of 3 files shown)


Joel-Dahne:
While neat it's quite odd. I wouldn't expect `getindex` to convert to `Float64` and if I encountered this in code not knowing about it I would be very confused. It feels a lot more clear to write `Float64(m)` or `convert(Float64, m)`.

saschatimme:
Yeah, I also wondered whether it's too special. I agree that writing `Float64(m)` look much clearer. I will add instead constructors and a convert method.

saschatimme:
@Joel-Dahne this is now the cleaner version

kalmarek (changes requested):
I think just the second line is necessary, `convert` for number types falls back to the default constructor:
```julia
julia> convert(Float64, Arf(0.1))
ERROR: MethodError: no method matching Float64(::Arf)
Closest candidates are:
  Float64(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:200
  Float64(::T) where T<:Number at boot.jl:716
  Float64(::UInt32) at float.jl:66
  ...
Stacktrace:
 [1] convert(::Type{Float64}, ::Arf) at ./number.jl:7
 [2] top-level scope at REPL[10]:1
```
also, maybe I'm mistaken but `get` doesn't have a method for `Arf`??
```julia
julia> methods(Arblib.get)
# 2 methods for generic function "get":
[1] get(x::Union{Ptr{Arblib.mag_struct}, Arblib.Mag, Arblib.mag_struct}) in Arblib at /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:266
[2] get(x::Union{Ptr{Arblib.arb_struct}, Arb, ArbRef, Arblib.arb_struct}, n::Integer, flags::Unsigned) in Arblib at /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:266
```
I know these are one-liners, but please add tests :)

saschatimme:
Sorry, I guess I am too sleep deprived today 😴 , I will clean this up tomorrow and add some proper tests. But Arf has a get method it is just commented out for unclear reasons.

Joel-Dahne:
The reason `arf_get_d` is commented out is that it clashes with `arf_get_si`, they both get the signature `get(::Arf, ::arb_rnd)`. I believe there are a few other cases like this and at the moment we have not decided on a way to handle them. For now I would say just manually implement the `ccall` for `convert`.

saschatimme:
I hope this is now in a better shape:
* `Float64` constructors from `Arf` and `Mag`
* `Int` constructor from `Arf`
* The constructors from `Arf` have a manual `ccall` to void the ambiguities.
* Proper tests

kalmarek (approved):

PR #56, opened 2020-09-02
Enh/eigen

kalmarek:
adds full support for eigenvalues computation;

Arblib allows passing some of the arguments as `C_NULL`, so I needed to abuse our arbcall macro ;) worked flawlessly ;)

In general this is the interface to all functions (whenever makes sense, e.g. for enclosures it doesn't)
```julia
Arblib.eigen_simple(A::AcbMatrix, side=:right) # compute eivenvalues AND (right) eigenvectors of A
Arblib.eigen_simple!(λs::AcbVector, A::AcbMatrix) # compute eigenvalues and store them in λs
Arblib.eigen_simple!(λs::AcbVector, eigvecs::AcbMatrix, A::AcbMatrix, side=:right) # compute and store eigenvalues and eigenvectors in λs and eigvecs. whether these are stored as columns or rows (left or right) is decided by side kwarg
Arblib.eigen_simple!(λs::AcbVector, eigvecs::AcbMatrix, A::AcbMatrix, λ_approx::AcbVector, R_eigvecs_approx::AcbMatrix, side=:right) # the same as above using the approximations provided by λ_approx and R_eigvecs_approx
```
the standard `arbcall` provides also
```julia
Arblib.eigen_simple!(λs::AcbVector, Lev::AcbMatrix, Rev::AcbMatrix, A::AcbMatrix, λ_approx::AcbVector, R_eigvecs_approx::AcbMatrix) # compute both left and right eigenvectors
```
some more kwargs are available for `approx_eig_qr`

More remarks:
 * `eig_simple_rump!(..., side=:left)` segfaults in Arblib; I'm not sure if this is a bug in the library, or a problem with our wrapper
 * at the moment tests print something like
```julia
Arblib.get(ε) = 2.082432852188159e-12
Arblib.get(twice_ε) = 4.164865717928845e-12
[abs(λa - λ) for λ = λs] = Arb[[+/- 1.62e-12], [1.64814713300 +/- 6.35e-12], [1.90905100186 +/- 6.07e-12]]
[abs(λa - λ) for λ = λs] = Arb[[1.64814713299 +/- 5.85e-12], [+/- 2.55e-12], [0.31315429338 +/- 3.91e-12]]
[abs(λa - λ) for λ = λs] = Arb[[1.90905100187 +/- 6.04e-12], [0.31315429338 +/- 3.91e-12], [+/- 2.55e-12]]
```
I think there is something wrong with `eig_global_enclosure!`: `ε` is supposed to provide upper bound on the distance of approximate eigenvalues to the actual eigenvalues. Clearly it doesn't, so I use `twice_ε`, but this may sometimes fail (matrix is random)
 * at the moment the signatures are simplistic; I will broaden them after @saschatimme Matrix pull is done, we're happy with this interface and the above problems are solved


saschatimme:
I think we can now merge this after rebasing :)

kalmarek:
ok, so now I have a dilemma: In the initial commit I could bounds-check all calls, since the arguments were honest `AcbMatrix/AcbVector`; Now If I'd like to use `AcbMatrixLike`, I can't do it (there is no `size(::Ptr{acb_mat_struct})` and I don't think there should be one). Decision:
* allow calling with `Union{AcbMatrix, AcbRefMatrix}` only, or
* keep those checking behind `@boundscheck`, so calling the low-level interface with `Ptr{acb_mat_struct}`s is possible with `@inbounds`;

The latter is kind of a hack, but a workable solution. Also note that the higher level functions (calling e.g. `similar`) will be defined with `Union{AcbMatrix, AcbRefMatrix}` only

saschatimme:
Why can't you define
```julia
size(::Ptr{acb_mat_struct}) = size(unsafe_load(acb_mat_struct))
```
? I mean this is unsafe, but `Ptr` life is anyways dangerous?

Joel-Dahne:
I think it makes perfect sense to only allow calling with `Union{AcbMatrix, AcbRefMatrix}`. They are the intended interface for the package, `arb_mat_struct` is an implementation detail. If you for some reason want to do it with a `arb_mat_struct` you can always fall back to using the low-level interface or trivially wrap it in an `AcbMatrix` with minimal overhead compared to the rest of the computations.

I agree that `size(::Ptr{acb_mat_struct})` should not be defined.

kalmarek:
I think we should be actually phasing out the usage of `Ptr{...}` since we have `Ref` types which will be just as good

Joel-Dahne:
Good point! So then you can at least skip it for now!

kalmarek:
I feel like those tests may fail occasionally, so maybe I should hardcode a single random example?

Joel-Dahne:
Probably a good idea. We only have to test correctness of the interface, not the algorithm.

kalmarek:
6 tests are marked as broken, since the `eig_global_enclosure` doesn't produce what the Arb docs say;
I'm not sure if this is a problem with Arb docs, or not:
http://arblib.org/acb_mat.html#c.acb_mat_eig_global_enclosure
states that the enclosure is given in absolute value (euclidean norm), however:

```julia
let
    A = rand(3,3)
    B = rand(3,3)
    
    M = AcbMatrix(A + B * im, prec = 64)
    
    tol = 1e-12
    λ_approx, R_approx = Arblib.approx_eig_qr(M, tol = tol)
    
    ε = Arblib.eig_global_enclosure!(Arblib.Mag(), M, λ_approx, R_approx)
    
    λs = Arblib.eig_simple!(similar(M, size(M,1)), M, λ_approx, R_approx) # too keep the order
    
    k = 10
    @info Arblib.get(ε)
    @info abs.(λs - λ_approx) .< k*Arb(Arf(ε))
    @info real.(λs - λ_approx) .< k*Arb(Arf(ε))
    @info imag.(λs - λ_approx) .< k*Arb(Arf(ε))
end
```
even with `k=10` this prints various numbers of `0` and `1` s...

saschatimme (approved):
I think this looks good for now, we can always iterate later on.

PR #57, opened 2020-09-03
Arbcall refactor

### LINKED ISSUE #55: len detection in arbcall is broken
this commit broke it:
https://github.com/kalmarek/Arblib.jl/commit/57fff067ab8eab62bf1b0e5b45f5d15fc8582aea
@saschatimme can you explain what was the reason to remove `Ptr{arb_struct}` from `jltype(ca::Carg{ArbVector})` (and the same for `AcbVector`? (I really need to undust my memory about the function of all of those `Unions`...)

In any case https://github.com/kalmarek/Arblib.jl/blob/dc8743597a5ea1d2c6eeb76d360db1d7c706516e/src/arbcall.jl#L198
needs to be kept in sync with `jltype`

also: none of the tests caught it; we need to add
```julia
            (
                "int _acb_vec_is_zero(acb_srcptr vec, slong len)",
                [
                    :(vec::$(Union{AcbVector, Arblib.acb_vec_struct})),
                ],
                [
                    Expr(:kw, :(len::Integer), :(length(vec)))
                ]
            )
```
to `jlargs` testset.

saschatimme:
This refactors the arbcall logic to fix #55 and to hopefully avoid things like this happening in the future.
This also adds some more features:
* Defines constants for the type unions we consider equivalent, e.g., `const ArbLike = Union{Arb,ArbRef,Ptr{arb_struct}}`.
* Overloads the show method for these constants to get more readable output:
```julia
# before
  Closest candidates are:
    add!(::Union{ArbVector, Arblib.arb_vec_struct}, ::Union{ArbVector, Arblib.arb_vec_struct}, ::Union{ArbVector, Arblib.arb_vec_struct}; n, prec) at /Users/sascha/code/Arblib.jl/src/arbcall.jl:275

# now
  Closest candidates are:
    add!(::ArbVectorLike, ::ArbVectorLike, ::ArbVectorLike; n, prec) at /Users/sascha/code/Arblib.jl/src/arbcall.jl:275
```
* The detection of length arguments is better now. If the argument is an `Int` and the previous argument `ArbVectorLike` or `AcbVectorLike` then the name of this argument is moved to be a kwarg, but only if the name is `n` or starts with `len`. 
* I also enabled in a separate commit  the possibility to automatically also generate the true arb signature, i.e., without moving any positional argument to a keyword argument.

This also probably needs some more tests, but this not a task for 10pm :D


kalmarek (approved):

kalmarek (approved):

kalmarek:
I'm happy with the changes, but @Joel-Dahne have a look at this as well

saschatimme:
@Joel-Dahne @kalmarek  and myself are both happy with the changes and think that it's a good idea to always also generated the original signature as an escape hatch if our replacement logic is non-sensical. What do you think?

src/arbcall.jl
```diff
@@ -163,76 +168,66 @@ function jlfname(
...
+            # Automatic detection of length arguments for vectors
+        elseif (startswith(string(name(carg)), "len") || carg == Carg{Clong}(:n, false)) &&
+               rawtype(cargs[i-1]) ∈ (ArbVector, AcbVector) &&
+               name(carg) ∉ len_keywords
+
+            vec_name = name(cargs[i-1])
+            push!(kwargs, Expr(:kw, :($(name(carg))::Integer), :(length($vec_name))))
+            push!(len_keywords, name(carg))
+
+        else
```
Joel-Dahne, line 223: Could we factor this out as a separate method? Mainly to make it easier to test independently since this is quite delicate.

src/types.jl
```diff
@@ -235,6 +220,30 @@ cstruct(x::AcbRef) = x.acb_ptr
...
+    AcbRef,
+    ArbVector,
+    AcbVector,
+    ArbMatrix,
+    AcbMatrix,
+    ArbRefVector,
+    AcbRefVector,
+    ArbRefMatrix,
+    AcbRefMatrix,
+}
```
Joel-Dahne, line 245: Minor thing, but can we either keep the `Ref` versions just under the default version or have all of them in the end, to make it easier to see directly what is going on.

Joel-Dahne (commented):
One issue with the automatic length is that you can still pass in two vectors with different lengths and crash Julia. So all of these methods should be considered unsafe. If we ever add exported versions of them we need to add proper bounds checks when wrapping them.

Joel-Dahne:
I did a very quick review of the code. In general I think it looks good, less code now. I think generating the original signature as well is reasonable! If we merge this then I can update #32 before merging that.

src/types.jl
saschatimme: Done

src/arbcall.jl
saschatimme: I factored it out and added some tests

PR #59, opened 2020-09-03
use ispredicate(::Arbfunction) to return Bool from arbcall

kalmarek:
This adds just predicate detection and uses it to turn the returned `int` to `Bool`. Arblib returns "success", hence non-zero output can be interpreted as `true`

### DIFF
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -184,7 +184,7 @@ function ispredicate(af::Arbfunction)
            (
                any(s -> startswith(string(jlfname(af)), s), ("is_",)) ||
                any(
-                   s -> contains(string(jlfname(af)), s),
+                   s -> occursin(s, string(jlfname(af))),
                    ("_is_", "contains", "can_", "check_", "validate_"),
                ) ||
                any(==(jlfname(af)), (:eq, :ne, :lt, :le, :gt, :ge, :overlaps, :equal))
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -178,6 +178,15 @@ function extract_length_argument!(kwargs, len_keywords, carg, prev_carg)
     push!(kwargs, Expr(:kw, :($(name(carg))::Integer), :(length($vec_name))))
     push!(len_keywords, name(carg))
 end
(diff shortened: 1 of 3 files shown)


src/arbcall.jl
```diff
@@ -168,6 +168,16 @@ function jlfname(
     return jlfname(arbfname(af), prefixes, suffixes, inplace = inplace)
 end
 
+function ispredicate(af::Arbfunction)
+    return isconst(first(arguments(af))) &&
+           returntype(af) == Cint &&
+           (
+               any(s -> startswith(string(jlfname(af)), s), ("is_", )) ||
+               any(s ->   contains(string(jlfname(af)), s), ("_is_", "contains", "can_", "check_", "validate_")) ||
```
Joel-Dahne, line 176: There is a very annoying extra space here...

Joel-Dahne (approved):
This looks good! Would be good to also add some simple tests, should be easy since `ispredicate` is self contained.

saschatimme:
@kalmarek In which branch do you want to merge this by now? Shouldn't this be master?

kalmarek:
@Joel-Dahne I know it's post-factum, but could you produce a list with your (canonical ;) version of arb docs which functions are detected as predicates? I made this list locally and everything looked ok, but you never know ;)

Joel-Dahne:
> @Joel-Dahne I know it's post-factum, but could you produce a list with your (canonical ;) version of arb docs which functions are detected as predicates? I made this list locally and everything looked ok, but you never know ;)

I looked through the functions detected as predicates and it seems fully correct! However it does miss some functions which do return a boolean, for example `arb_mat_solve`, but this is probably fine. I can give a precise list as well, but that probably makes more sense to do after #32 is merged since much more functions will be parsed.

kalmarek:
I think that the return value of `arb_mat_solve` (also of e.g. `eig_simple`) is more of an error code than a `Bool`, but we can discuss this and change it later. Indeed the `ispredicate` would become much simpler, since there are only a few exceptions to the `Cint → Bool` translation upon return (e.g. `*_cmp` function).

PR #61, opened 2020-09-04
Add MagRef and ArfRef

saschatimme:
This superseeds #52 since the rebasing of #52 onto master was basically not possible. I decided to also directly add a `MagRef` type since we need it anyways. With this PR is not now possible to convert an `Acb` to a `ComplexF64` (by using the midpoint of `Acb`) without any allocations.

### DIFF
--- a/src/predicates.jl
+++ b/src/predicates.jl
@@ -94,20 +94,15 @@ end
 #Base.:(<=)(x::Arf, y::Arf) = !isnan(x) && !isnan(y) && cmp(x, y) <= 0
 #Base.isequal(x::Arf, y::Arf) = !iszero(is_equal(x, y))
 
-for ArbT in (MagLike, ArfLike, ArbLike, AcbLike)
+for ArbT in (ArfLike, ArbLike, AcbLike)
     @eval begin
         Base.isequal(y::$ArbT, x::$ArbT) = !iszero(equal(x, y))
-    end
-
-    ArbT == MagLike && continue
-
-    # Comparison of non-floating point values should use ==
-    @eval begin
+        # Comparison of non-floating point values should use ==
         Base.:(==)(y::Integer, x::$ArbT) = !iszero(equal(x, y))
         Base.:(==)(x::$ArbT, y::Integer) = !iszero(equal(x, y))
     end
 end
-
+Base.:(==)(x::MagLike, y::MagLike) = !iszero(equal(x, y))
 Base.isless(x::MagLike, y::MagLike) = cmp(x, y) < 0
 Base.:(<)(x::MagLike, y::MagLike) = cmp(x, y) < 0
 Base.:(<=)(x::MagLike, y::MagLike) = cmp(x, y) <= 0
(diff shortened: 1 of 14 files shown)


src/predicates.jl
```diff
@@ -92,40 +92,36 @@ end
 #Base.:(<=)(x::Arf, y::Arf) = !isnan(x) && !isnan(y) && cmp(x, y) <= 0
 #Base.isequal(x::Arf, y::Arf) = !iszero(is_equal(x, y))
```
Joel-Dahne, line 93: Can you remove these and the other commented out part above now that you are at it?

src/types.jl
```diff
@@ -244,20 +274,37 @@ const ArbTypes = Union{
...
 end
-function Base.Int(
-    x::Union{Arf,Ptr{arf_struct}};
-    rnd::Union{arb_rnd,RoundingMode} = RoundNearest,
-)
+function Base.Int(x::ArfLike; rnd::Union{arb_rnd,RoundingMode} = RoundNearest)
     ccall(@libarb(arf_get_si), Clong, (Ref{arf_struct}, arb_rnd), x, rnd)
 end
+Base.Float64(x::ArbLike) = Float64(midref(x))
+Base.ComplexF64(z::AcbLike) = Complex(Float64(realref(z)), Float64(imagref(z)))
```
Joel-Dahne, line 310: These might make more sense to have in a `conversion.jl` file. But that can be fixed at a later point.

arb_mat.rst
```diff
@@ -0,0 +1,759 @@
+.. _arb-mat:
```
Joel-Dahne, line 1: This file should probably not be here?

Joel-Dahne (changes requested):
Looks good! This will make the interface a lot more uniform. The only real problem is the file that should probably not be there.

arb_mat.rst
saschatimme: Dang, thanks for catching this

src/predicates.jl
saschatimme: Sure, done.

src/types.jl
saschatimme: Make sense, but this probably best to group together with maybe some other regrouping.

Joel-Dahne (approved):
Good job!

arb_mat.rst
Joel-Dahne: At first I thought you had started to add documentation, which would have been great 😂

Joel-Dahne:
That merge was a miss-click, keep it or rebase over it if you prefer...

src/types.jl
```diff
@@ -244,20 +274,37 @@ const ArbTypes = Union{
...
+end
+
+function realref(z::Union{Acb,AcbRef}; prec = _precision(z))
+    real_ptr = ccall(@libarb(acb_real_ptr), Ptr{arb_struct}, (Ref{acb_struct},), z)
+    ArbRef(real_ptr, prec, parentstruct(z))
+end
+function imagref(z::Union{Acb,AcbRef}; prec = _precision(z))
+    real_ptr = ccall(@libarb(acb_imag_ptr), Ptr{arb_struct}, (Ref{acb_struct},), z)
+    ArbRef(real_ptr, prec, parentstruct(z))
+end
```
kalmarek, line 300: use `prec=precision(z)`

kalmarek (changes requested):
this looks great!

kalmarek (approved):
looking great!

PR #66, opened 2020-09-09
add initial version of integration

kalmarek:
I took the old code I've written wrapping integration and dumped it here;

at the moment:
* it's 2-orders of magnitude slower than Quadgk on the `sin` example:
```julia
julia> @btime Arblib.integrate($(Arblib.sin!), $(Acb(0)), $(Acb(1)), prec=53)
  22.574 μs (54 allocations: 2.14 KiB)
[0.459697694131860 +/- 6.21e-16]

julia> using QuadGK

julia> @btime quadgk($sin, 0, 1)
  248.713 ns (5 allocations: 144 bytes)
(0.4596976941318603, 5.551115123125783e-17)
```
* the function to be integrated is required to be of the signature `(res::Ptr{acb_struct}, x::Ptr{acb_struct}; prec) → (...)` and must modify `res`
* a slightly more user-firendly way of doing this goes through this:
```julia
function _acb_calc_func2!(
    out::Ptr{acb_struct},
    inp::Ptr{acb_struct},
    param::Ptr{Cvoid}, # pointer to the actual function
    order::Cint,
    prec::Cint,
)
    @assert iszero(order) || isone(order) # ← we'd need to verify holomorphicity
    x = Acb(unsafe_load(inp), prec=prec)
    # x = inp
    f = unsafe_pointer_to_objref(param)
    # @debug "Evaluating at" x
    set!(out, f(x))
    return zero(Cint)
end
```
allocates twice as much (at ~10% slowdown) but allows to write code like this:
```
julia> @btime Arblib.integrate($sin, $(Acb(0)), $(Acb(1)), prec=53)
  24.860 μs (102 allocations: 7.89 KiB)
[0.459697694131860 +/- 6.21e-16]
```
* tests are to be written

### DIFF
--- a/src/calc_integrate.jl
+++ b/src/calc_integrate.jl
@@ -2,22 +2,26 @@ function _acb_calc_func!(
     out::Ptr{acb_struct},
     inp::Ptr{acb_struct},
     param::Ptr{Cvoid}, # pointer to the actual function
-    order::Cint,
-    prec::Cint,
+    order::Clong,
+    prec::Clong,
 )
     @assert iszero(order) || isone(order) # ← we'd need to verify holomorphicity
     # x = Acb(unsafe_load(inp), prec=prec)
-    x = inp
-    f = unsafe_pointer_to_objref(param)
-    # @debug "Evaluating at" x
-    f(out, x, prec = prec)
+    # TODO: Make Acb allow Nothing as parent
+    tmp = AcbVector(1)
+    x = AcbRef(inp, cstruct(tmp), prec = prec)
+    res = AcbRef(out, cstruct(tmp), prec = prec)
+    f! = unsafe_pointer_to_objref(param)
+
+    f!(res, x, analytic = isone(order), prec = prec)
+
     return zero(Cint)
 end
 
 _acb_calc_cfunc!() = @cfunction(
     _acb_calc_func!,
     Cint,
-    (Ptr{acb_struct}, Ptr{acb_struct}, Ptr{Cvoid}, Cint, Cint)
+    (Ptr{acb_struct}, Ptr{acb_struct}, Ptr{Cvoid}, Clong, Clong)
 )
 
 function calc_integrate!(
@@ -31,7 +35,6 @@ function calc_integrate!(
     options::calc_integrate_opt_struct,
     prec::Clong,
 )
-    @info "ccalling in calc_integrate!"
     return ccall(
         @libarb(acb_calc_integrate),
         Cint,
@@ -95,22 +98,27 @@ function calc_integrate!(
     )
 end
 
+"""
+    _integrate!(res, f, a, b; prec, rel_goal, abs_goal, opts)
 
-function integrate!(
+Internal function for integration. Assumes that `f!` is of the type
+`f!(out::AcbRef, inp::AcbRef; analytic::Bool, prec::Integer)` and
+that all other parameters are set to valid values.
+"""
+function _integrate!(
+    f!,
     res::AcbLike,
-    f,
     a::AcbLike,
     b::AcbLike;
-    prec = max(_precision(a), _precision(b)),
-    rel_goal = prec,
+    prec::Integer = precision(res),
+    rel_goal::Integer = prec,
     abs_tol::MagLike = set_ui_2exp!(Arblib.Mag(), one(UInt), -prec),
     opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
 )
-
-    return calc_integrate!(
+    status = calc_integrate!(
         res,
         _acb_calc_cfunc!(),
-        f,
+        f!,
         a,
         b,
         rel_goal,
@@ -118,7 +126,97 @@ function integrate!(
         opts, # passing C_NULL uses the default options
         prec,
     )
+
+    # status:
+    # ARB_CALC_SUCCESS = 0
+    # ARB_CALC_NO_CONVERGENCE = 2
+    status == 2 && @warn "integration did not converge"
+
+    return res
 end
+
+"""
+    integrate!(f!, res::Acb, a::Acb, b::Acb;
+        prec = precision(res),
+        rtol=0.0,
+        atol=2.0^-prec,
+        opts::Union{acb_calc_integrate_opt_struct, Ptr{Cvoid}} = C_NULL)
+
+Like [`integrate`](@ref), but make use of in-place operations. In
+particular, there are three differences from `integrate`:
+
+1. The function `f!` should be of the form `f!(y, x) = set!(y, f(x))`.
+   That is, it writes the return value of the integand `f(x)` in-place
+   into its first argument `y`. (The return value of `f!` is ignored).
+
+2. `res` is set to the result.
+
+3. The default precision is taken from `res` instead of from `a` and
+   `b`.
+"""
+function integrate!(
+    f!,
+    res::Union{Acb,AcbRef},
+    a::Union{Acb,AcbRef},
+    b::Union{Acb,AcbRef};
+    check_analytic::Bool = false,
+    take_prec::Bool = false,
+    prec::Integer = precision(res),
+    rtol = 0.0,
+    atol = set_ui_2exp!(Mag(), one(UInt), -prec),
+    opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
+)
+    # rel_goal = r where rel_tol ~2^-r
+    rel_goal = iszero(rtol) ? prec : rel_goal = -floor(Int, log2(atol))
+
+    if !check_analytic && !take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out)
+    elseif !check_analytic && take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out, prec = prec)
+    elseif check_analytic && !take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out, analytic = analytic)
+    else
+        g! = f!
+    end
+
+    return _integrate!(
+        g!,
+        res,
+        a,
+        b,
+        prec = prec,
+        rel_goal = rel_goal,
+        abs_tol = convert(Mag, atol),
+        opts = opts,
+    )
+end
+
+function integrate!(
+    f!,
+    res::Union{Acb,AcbRef},
+    a::Number,
+    b::Number;
+    check_analytic::Bool = false,
+    take_prec::Bool = false,
+    prec::Integer = max(_precision(a), _precision(b)),
+    rtol = 0.0,
+    atol = set_ui_2exp!(Mag(), one(UInt), -prec),
+    opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
+)
+    return integrate!(
+        f!,
+        res,
+        Acb(a, prec = prec),
+        Acb(b, prec = prec),
+        check_analytic = check_analytic,
+        take_prec = take_prec,
+        prec = prec,
+        rtol = rtol,
+        atol = atol,
+        opts = opts,
+    )
+end
+
 
 """
     integrate(f, a::Number, b::Number;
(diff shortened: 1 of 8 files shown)


Joel-Dahne:
I think we should definitely allow the more user friendly version as well. In your example it adds a 10% overhead but this is likely much lower for a more costly to compute function than `sin`. `QuadGK` also allows both inplace and non-inplace integrands, so we can do something similar to them.

Does your current implementation allow using the `order` at all? We should handle that in some way as well.

Regarding being a lot slower than `QuadGK` this is purely because hardware floating points are about 2 orders of magnitude faster than arbitrary precision ones. Compare
```julia
julia> @btime Arblib.integrate($(Arblib.sin!), $(Acb(0)), $(Acb(1)), prec=53)
  26.276 μs (54 allocations: 2.14 KiB)
[0.459697694131860 +/- 6.21e-16]

julia> setprecision(53)
53

julia> @btime quadgk($sin, $(big(0)), $(big(1)))
  30.067 μs (310 allocations: 9.41 KiB)
(0.45969769413186035, 5.5511151231257827e-17)
```

kalmarek:
yes, I even have an idea for traits code to handle user-friendly/fast versions;

You tell me what shall I do with `order` argument ;)

Joel-Dahne:
I have some ideas for how to handle in-place and out-of-place version and I'm working on some code for it. Should I push here or somewhere else when I get it done?

kalmarek:
I was just planning to dispatch on the type `Inplace`/`Outofplace`, but sure, go ahead!

Joel-Dahne:
I realised it's probably better if I write here first. I was also working on how to handle in-place vs out-of-place functions, the way I see it we have three alternatives:

The first is to do what `QuadGK.jl` does and have to separate methods, `quadgk(f, a, b)` and `quadgk!(f!, res, a, b)` where the first expects an out-of-place function and the latter an in-place function. 

The second is to do what `DifferentialEquations.jl` does and try to determine if the given function is in-place or out-of-place, I'm still not completely sure how they do it, but also support specifying it with a keyword argument.

The third is what you propose, to dispatch on some `Inplace`/`Outofplace` type.

I sort of like the first option, it limits the options a bit but it's not unreasonable to have them specify the `res` argument if they give an in-place function. What do you think?

Joel-Dahne:
Finally got around to finish most of the work I started. There's still things to be done, but here is a first version at least. There are two main methods, `integrate` and `integrate!`, the first one expects an out-of-place function and the second one an in-place function. They also both accepts the keyword arguments `take_prec` and `check_analytic`. If `take_prec` is true then the function is given `prec` as a keyword argument. If `check_analytic` is true then the function is given `analytic::Bool` as a keyword argument. There is still some documentation to add about these two options. 

What I'm slightly worried about is if there are any performance implications coming from how the final function is determined. I have not found any and likely it would be overshadowed by the cost of the numerical computation. Still might be worth to take another look at.

Joel-Dahne:
I decided to try and finish this instead of starting with other things. I have rebased on master and added some more documentation. I want to add a few examples to the documentation as well before finishing it.

src/calc_integrate.jl
```diff
@@ -0,0 +1,325 @@
...
+
+    if !check_analytic && !take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out)
+    elseif !check_analytic && take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out, prec = prec)
+    elseif check_analytic && !take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out, analytic = analytic)
+    else
+        g! = f!
+    end
```
Joel-Dahne, line 180: I'm also slightly worried about this construction of `g!` having performance implications. I don't think so, but I'm not sure.

src/calc_integrate.jl
```diff
@@ -0,0 +1,325 @@
...
+    a::Union{Acb,AcbRef},
+    b::Union{Acb,AcbRef};
+    check_analytic::Bool = false,
+    take_prec::Bool = false,
+    prec = max(precision(a), precision(b)),
+    rtol = 0.0,
+    atol = set_ui_2exp!(Mag(), one(UInt), -prec),
+    opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
+)
+    f! = (out, inp; kwargs...) -> set!(out, f(inp; kwargs...))
```
Joel-Dahne, line 286: Also here I'm not sure about the performance implications.

src/calc_integrate.jl
```diff
@@ -0,0 +1,325 @@
...
+)
+    # rtol ≈ 2^(-rel_goal)
+    rel_goal = iszero(rtol) ? prec : max(-floor(Int, log2(atol)), 0)
+
+    if !check_analytic && !take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out)
+    elseif !check_analytic && take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out, prec = prec)
+    elseif check_analytic && !take_prec
+        g! = (inp, out; analytic, prec) -> f!(inp, out, analytic = analytic)
```
kalmarek, line 177: change order of `inp`, `out` ;) I'd like to have these named `f!(res, x, ...)`, but that might be confusing with the `res` argument (though I believe perfectly valid).

I'm not sure how expensive defining (and compiling) those anonymous functions will be;
Another solution would be to define
```julia
struct CalcFunction{Analytic, TakePrec, F}
    func::F
    CalcFunction(f, inplace::Val{true}; check_analytic::Bool=false, take_prec=false) = 
        new{check_analytic, take_prec}(f)
    CalcFunction(f, inplace::Val{false}; check_analytic::Bool=false, take_prec=false) = 
        new{check_analytic, take_prec}(res, x) -> set!(res, f(x)))
end

(F::CalcFunction{false,false})(res, x; analytic, prec) = F.func(res, x)
(F::CalcFunction{true,false})(res, x; analytic, prec) = F.func(res, x, analytic=analytic)
....
```
and pass this guy directly as `param`. I'm not sure if this would actually be more efficient...
I'd be actually interested if deparametrizing `func` would actually change performance ;)

src/calc_integrate.jl
```diff
@@ -0,0 +1,325 @@
+function _acb_calc_func!(
+    out::Ptr{acb_struct},
+    inp::Ptr{acb_struct},
+    param::Ptr{Cvoid}, # pointer to the actual function
+    order::Clong,
+    prec::Clong,
+)
+    @assert iszero(order) || isone(order)
+    x = AcbRef(inp, nothing, prec = prec)
+    res = AcbRef(out, nothing, prec = prec)
```
kalmarek, line 10: does using `Acb` here allocate?

src/calc_integrate.jl
```diff
@@ -0,0 +1,325 @@
...
+2. `res` is set to the result.
+
+3. The default precision is taken from `res` instead of from `a` and
+   `b`.
+"""
+function integrate!(
+    f!,
+    res::Union{Acb,AcbRef},
+    a::Union{Acb,AcbRef},
+    b::Union{Acb,AcbRef};
```
kalmarek, line 161: (after rebase) `AcbOrRef`

src/calc_integrate.jl
kalmarek: there won't be any, with just two of them, it'll be compiled once for each `f`

kalmarek (commented):
looks good! maybe we can explore using `CalcFunction` for a while?

src/calc_integrate.jl
kalmarek: for sure it does :) silly me

src/calc_integrate.jl
kalmarek: So I did:
```julia
struct _CalcFunction{Analytic, Precision, F} <: Function
    func::F
    _CalcFunction(f; check_analytic::Bool=false, take_prec::Bool=false) = new{check_analytic, take_prec, typeof(f)}(f)
end

(F::_CalcFunction{false, false})(res::AcbLike, x::AcbLike; analytic, prec) =
    F.func(res, x)
(F::_CalcFunction{true, false})(res::AcbLike, x::AcbLike; analytic, prec) =
    F.func(res, x, analytic=analytic)
(F::_CalcFunction{false, true})(res::AcbLike, x::AcbLike; analytic, prec) =
    F.func(res, x, prec=prec)
(F::_CalcFunction{true, true})(res::AcbLike, x::AcbLike; analytic, prec) =
    F.func(res, x, analytic=analytic, prec=prec)

function integrate2!(
    f!,
    res::Union{Acb,AcbRef},
    a::Union{Acb,AcbRef},
    b::Union{Acb,AcbRef};
    check_analytic::Bool = false,
    take_prec::Bool = false,
    prec::Integer = precision(res),
    rtol = 0.0,
    atol = set_ui_2exp!(Mag(), one(UInt), -prec),
    opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
)
    # rtol ≈ 2^(-rel_goal)
    rel_goal = iszero(rtol) ? prec : max(-floor(Int, log2(atol)), 0)

    g! = _CalcFunction(f!, check_analytic=check_analytic, take_prec=take_prec)

    return _integrate!(
        g!,
        res,
        a,
        b,
        prec = prec,
        rel_goal = rel_goal,
        abs_tol = convert(Mag, atol),
        opts = opts,
    )
end
```

benchmark:
```julia
using Arblib

using Test
using BenchmarkTools
using QuadGK

let prec = 256
    a = Acb(0, prec = prec)
    b = Acb(1, prec = prec)
    out = Acb(a)
    @btime Arblib.integrate!($(Arblib.sin!), $(out), $(a), $(b))
    @btime Arblib.integrate2!($(Arblib.sin!), $(out), $(a), $(b))
    setprecision(BigFloat, prec) do
        @btime quadgk($sin, $(big(0.0)), $(big(1.0)))
    end
end;
```
gives:
```
# first time:
  78.148 μs (142 allocations: 5.67 KiB)
  78.828 μs (142 allocations: 5.67 KiB)
  3.678 ms (17660 allocations: 805.05 KiB)
# second time
  78.237 μs (138 allocations: 5.55 KiB)
  78.685 μs (138 allocations: 5.55 KiB)
  3.709 ms (17660 allocations: 805.05 KiB)
```

There is just 2µs impact when `func::Any`; not bad :)

kalmarek: in the same situation
```julia
@btime Arblib.integrate($sin, $a, $b)
```
takes
```
82.385 μs (223 allocations: 12.08 KiB)
```
this looks really great!

Joel-Dahne: Okay so it seems like we are good with the current version? I can change the arguments to `g!`. The mimic the ones for Arb at the moment, but it's probably better to follow Julia standards.

src/ref.jl
```diff
@@ -19,7 +19,7 @@ function ArbRef(
 end
 function AcbRef(
     ptr::Ptr{acb_struct},
-    parent::Union{acb_vec_struct,acb_mat_struct};
+    parent::Union{Nothing,acb_vec_struct,acb_mat_struct};
```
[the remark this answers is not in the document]
Joel-Dahne: This is similar to how Julia Base does it. In `refpointer.jl` they specify that the `roots` as the call it should be either `Nothing` or `Any`. At the moment we use this in our code at any point. It's only there to keep the parent from getting GCed. In the case with the integration the parent struct is handled by the C-code so it doesn't have to be handled by Julia.

Not sure what you mean with leaving it uninitialised, just construct it with `new` and not set `parent` to a value?

src/calc_integrate.jl
Joel-Dahne: I realised I had also just named them in the wrong direction...

src/ref.jl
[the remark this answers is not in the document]
kalmarek: yes, partial initialization constructor; but as you say, it's internal so it doesn't matter that much; It's definitely not blocking from my perspective.

src/calc_integrate.jl
kalmarek: measuring the time to define an anonymous function:
```julia
let prec = 256
    a = Acb(0, prec = prec)
    b = Acb(1, prec = prec)
    out = Acb(a)
    @btime Arblib.integrate!($(Arblib.sin!), $(out), $(a), $(b))
    let g = (res, x; analytic, prec) -> Arblib.sin!(res, x)
        @btime Arblib._integrate!($g, $(out), $(a), $(b))
    end
    @btime Arblib.integrate2!($(Arblib.sin!), $(out), $(a), $(b))
```
```
  78.406 μs (138 allocations: 5.55 KiB)
  77.090 μs (138 allocations: 5.55 KiB)
  78.410 μs (138 allocations: 5.55 KiB)
```
it's between 1-1.5µs on my laptop;
I think this is good to go!

kalmarek:
It's pretty funny, I need to ask you to review since I opened the initial pull :P

Joel-Dahne:
I started writing some small examples to include in the documentation. I can hopefully finish that tomorrow and then we can merge!

Joel-Dahne (approved):
I think I have taken care of all our comments, so now I'm happy with this. Would be great if you could look over the updated documentation to see if it looks okay! Then you can merge it :smiley:

kalmarek:
@Joel-Dahne thanks for taking care of this!

PR #67, opened 2020-09-17
Enh/more methods

Joel-Dahne:
I have added functions from `acb_hypgeom.h` and `arb_hypgeom.h`. They both parsed cleanly, only 6 (of 212) functions failed to parse due to types we have not implemented (`acb_poly_struct *` and `fmpz_t`).

The files that are left that we are not parsing are
- `arb_fmpz_poly.h`
- `acb_dft.h`
- `acb_elliptic.h`
- `acb_modular.h`
- `dirichlet.h`
- `acb_dirichlet.h`
- `bernoulli.h`
- `hypgeom.h`
- `partitions.h`
- `arb_calc.h`
- `acb_calc.h`
- `fmpz_extras.h`
- `bool_mat.h`
- `dlog.h`
- `fmpr.h`

However most of these are not relevant at all (for example`bool_mat.h`) whereas others will required a mostly manual implementation (`arb_calc.h`).


kalmarek:
`acb_calc.h` should be covered by #66; other methods in the header are legacy ones

Joel-Dahne:
Should we merge this now or should I add parsing of some more header files first?

kalmarek:
since you're on `parse.jl` could you maybe add the two-letters codes suggested somewhere? like
`#ut arb_signature_here` for unsupported type, `#pf ...` for parse failure and maybe `#mo ...` for manual override?

It'd be great if each line with `#mo ...` could also get a short explanation why manual override is done (such information is forgotten in a week after merge ;) 

otherwise, I'm happy with merging

parse/parse.jl
```diff
@@ -73,14 +81,26 @@ function generate_file(title, sections)
                 str *= "arbcall\"" * s * "\"\n"
             catch e
                 if e isa Arblib.UnsupportedArgumentType
+                    push!(unsuppargs, e.key)
+                    num_unsupparg += 1
                     str *= "# arbcall\"" * s * "\"\n"
```
kalmarek, line 86:
```suggestion
                    str *= "#ut arbcall\"" * s * "\"\n"
```

parse/parse.jl
```diff
@@ -73,14 +81,26 @@ function generate_file(title, sections)
...
                 elseif e isa KeyError
+                    push!(keyerrors, e.key)
+                    num_keyerror += 1
                     str *= "# arbcall\"" * s * "\"\n"
```
kalmarek, line 90:
```suggestion
                    str *= "#pf arbcall\"" * s * "\"\n"
```

Joel-Dahne:
Good point! I'll take care of that!

Joel-Dahne:
How does this look?

kalmarek (approved):
wow, you went full throttle :1st_place_medal: !

PR #69, opened 2020-09-23
Cleanup types

Joel-Dahne:
I made some cleanup of the `types.jl` file, moving many of the constructors and simplifying some code. I want to move some of the methods defined in the bottom of the file, but I'm not sure exactly where or if I should do it in this pull request.

### DIFF
--- a/src/poly.jl
+++ b/src/poly.jl
@@ -60,7 +60,13 @@ Base.@propagate_inbounds function Base.setindex!(
 end
 
 # Constructors
-for TPoly in [:ArbPoly, :AcbPoly]
+for (TPoly, TPolyStruct) in [(:ArbPoly, :arb_poly_struct), (:AcbPoly, :acb_poly_struct)]
+    @eval function $TPoly(poly::$TPolyStruct; prec::Integer = DEFAULT_PRECISION[])
+        res = $TPoly(prec = prec)
+        set!(res, poly)
+        return res
+    end
+
     @eval function $TPoly(coeff; prec::Integer = _precision(coeff))
         poly = $TPoly(prec = prec)
         poly[0] = coeff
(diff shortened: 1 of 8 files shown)


src/poly.jl
```diff
@@ -76,7 +82,18 @@ for TPoly in [:ArbPoly, :AcbPoly]
     end
 end
 
-for (TSeries, T) in [(:ArbSeries, :Arb), (:AcbSeries, :Acb)]
+for (TSeries, TPolyStruct, T) in
```
kalmarek, line 85: does `T` occur in the body of the loop?

src/types.jl
```diff
@@ -270,9 +192,6 @@ for T in [:Arb, :Acb], A in [:Vector, :Matrix]
     end
 end
 
-const ArbLike = Union{Arb,ArbRef,Ptr{arb_struct},arb_struct}
-const AcbLike = Union{Acb,AcbRef,Ptr{acb_struct},acb_struct}
-
```
kalmarek, line 275: where did these go?

kalmarek (changes requested):

src/types.jl
Joel-Dahne: They had duplicate definitions, the are defined again further down in the file. I mentioned this in the commit message but I have realised that those messages mostly get lost because at least I basically only look at the total diff under "Files changed" where they are not shown...

src/poly.jl
Joel-Dahne: It does not! I must have removed that part in a earlier commit. Related to this:

Instead of specifying `TPolyStruct` I could do `TPolyStruct = cstructtype($TSeries)` inside the loop. This would probably be cleaner, but I was worried it would generate different code in the end, should maybe try it.

Joel-Dahne: Seems like it worked well, see my latest commit.

Joel-Dahne:
I also decided to factor out most of the `Ref` methods to a separate file, working on this now.

Joel-Dahne:
The nightly macOS-latest continues to fail for some likely unrelated reason. Is it possible to ignore that test for now? I don't really know how it works.

kalmarek:
I think we should have `fail-fast: false` in gh actions ci

saschatimme:
I pushed the change to the gh actions config such that always all tests are run. I also checked the latest mac os nightly and the test pass on my machine.

saschatimme (approved):

saschatimme:
@Joel-Dahne This looks great :) 

@kalmarek Are there still changes to be made from your point of view?

Joel-Dahne:
Great that you fixed the tests!

kalmarek (approved):

kalmarek:
No, I'm happy with the changes!

PR #71, opened 2020-10-13
Start readme

saschatimme:
I started a readme. @kalmarek @Joel-Dahne  please feel free to just push on this branch in case you find some time to add something :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@ From the Arb documentation:
 The following table indicates how Arb C-types can be translated to the Julia side.
 Note that all Julia structs additionally contain a `precision` field storing the precision
 used.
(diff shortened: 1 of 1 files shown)


Joel-Dahne:
What do you think about these changes?

saschatimme:
Looks good!

Joel-Dahne:
I couldn't get the example to run, fixed that. Also added type annotations to the method to be more clear about what it does. The example is good if we want to show that you can call Arb  directly, not very good if we want to show that it's convenient to use in general. Once we have added more convenience methods I guess we can create another example showing that.

Joel-Dahne (approved):
I think I'm happy with it in its current form. Anything more you want to add?

kalmarek:
If we're to release it fast, then let's merge this

PR #75, opened 2020-10-15
Fix setprecision for [Arb/Acb]Ref[Vector/Matrix]

saschatimme:


Joel-Dahne:
There is a slight problem with this and that's that `ArfLike,ArbLike,AcbLike` also include the underlying `cstructs` which do not have a precision set. For example `setprecision(Arb().arb, 128)` fails. I started doing some cleanup of the `constructors.jl` file and I realised also then that it would be nice to have a type-union which includes only the ref and non-ref version of a type, e.g. `Union{Arb,ArbRef}` since they are the only ones that can set a precision. I'm not sure if this is the best way though, it might make things harder to understand with so many different type-unions everywhere.

saschatimme:
I reverted the change to `setprecision` for Arb, Acb etc.

src/precision.jl
```diff
@@ -35,21 +35,16 @@ function Base.setprecision(::Type{<:ArbTypes}, precision::Integer)
...
-Base.setprecision(series::AcbSeries, precision::Integer) =
-    AcbSeries(series.acb_poly, degree(series); prec = precision)
-Base.setprecision(A::ArbMatrix, precision::Integer) = ArbMatrix(A.arb_mat, precision)
-Base.setprecision(A::AcbMatrix, precision::Integer) = AcbMatrix(A.acb_mat, precision)
+Base.setprecision(x::T, prec::Integer) where {T<:Union{Arf,ArfRef,Arb,ArbRef,Acb,AcbRef}} =
+    T(x, prec = prec)
+Base.setprecision(v::T, prec::Integer) where {T<:ArbVectorLike} = T(v.arb_vec, prec)
+Base.setprecision(v::T, prec::Integer) where {T<:AcbVectorLike} = T(v.acb_vec, prec)
+Base.setprecision(A::T, prec::Integer) where {T<:ArbMatrixLike} = T(A.arb_mat, prec)
+Base.setprecision(A::T, prec::Integer) where {T<:AcbMatrixLike} = T(A.acb_mat, prec)
```
Joel-Dahne, line 43: There is a similar problem here. `ArbVectorLike` includes `cstructtype(ArbVector)` for which the `setprecision` wont work. This seems like a theme we have encountered in several places now, you here and me when cleaning up the `constructors.jl`, that we need a union type for the ref and non-ref version. Maybe we could do something like
```julia
const CArbLike = Union{Arb,ArbRef,cstructtype(Arb),Ptr{cstructtype(Arb)}}
const ArbLike = Union{Arb,ArbRef}
```
to indicate that the first union is similar to `Arb` in the C sense and the second one in the Julia sense? This would correspond to how Julia uses `Cvoid`, `Cint` and such.

saschatimme: I like the idea. But I have some worries that the clean up get's out of hand. I would really really like to have a first release by the end of week since a project I am working on is somewhat stalled until we have an Arb release here...

Joel-Dahne: I understand! In that case I have no worries with doing a release after merging this (changing `ArbVectorLike` to `Union{ArbVector,ArbRefVector}` and same for the others) and also your `set!` for `Rational`. Then we can include my cleanup on the next release.

saschatimme: That sounds like a good plan :)

Joel-Dahne (approved):

PR #76, opened 2020-10-18
Cleanup constructors

Joel-Dahne:
I did some cleanup of `constructors.jl`. Mainly
- Rewriting to remove most of the `@eval`s
- Add some more constructors
- Add more test
- Add constructors for more `MathConstants`
- Move some code to `ref.jl`

Also, what do you think about adding `@inline _precision(x::BigFloat) = precision(x)`to `precision.jl`? I think that could make some of the logic simpler since we usually want to take the precision of a `BigFloat` into account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,11 +114,11 @@ add!(z::ArbLike, x::ArbLike, y::Int; prec = precision(z))
 
 Arb defines a number of functions for setting something to a specific
 value, for example `void arb_set_si(arb_t y, slong x)`. All of these
-are renamed to `set!` and really on multiple dispatch to choose the
-correct one. In addition to the ones defined in Arb there are a number
-of ones added in Arblib to make it more convenient to work with. For
-example there are setters for `Rational` and all irrationals defined
-in `Base.MathConstants`.
+are renamed to `set!` and rely on multiple dispatch to choose the
+correct one. In addition to the ones defined in Arb there is a number
+of methods of `set!` added in Arblib to make it more convenient to
+work with. For example there are setters for `Rational` and all
+irrationals defined in `Base.MathConstants`.
 
 Almost all of the constructors are simple wrappers around these
 setters. This means that it's usually more informative to look at the
--- a/README.md
+++ b/README.md
@@ -110,6 +110,31 @@ becomes
 add!(z::ArbLike, x::ArbLike, y::Int; prec = precision(z))
 ```
 
+## Constructors and setters
+
+Arb defines a number of functions for setting something to a specific
+value, for example `void arb_set_si(arb_t y, slong x)`. All of these
+are renamed to `set!` and really on multiple dispatch to choose the
+correct one. In addition to the ones defined in Arb there are a number
+of ones added in Arblib to make it more convenient to work with. For
+example there are setters for `Rational` and all irrationals defined
+in `Base.MathConstants`.
+
+Almost all of the constructors are simple wrappers around these
+setters. This means that it's usually more informative to look at the
+methods for `set!` than for say `Arb` to figure out what constructors
+exists. Both `Arb` and `Acb` are constructed in such a way that the
+result will always enclose the input.
+
+**Example:**
+``` julia
+x = Arblib.set!(Arb(), π)
+y = Arb(π)
+
+x = Arblib.set!(Arb(), 5//13)
+y = Arb(5//13)
+```
+
 ## Example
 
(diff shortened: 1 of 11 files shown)


saschatimme (commented):
Looks good for me. I jus found a couple of missing `Ref` variants.

> Also, what do you think about adding @inline _precision(x::BigFloat) = precision(x)to precision.jl

I think this makes sense.

Joel-Dahne:
Now I have added constructors for ref-types! I wanted to add `Mag(x::MagRef)` and `Mag(x::ArfRef)` as inner constructors to `Mag` but since `MagRef` and `ArfRef` are defined later I couldn't make that work.

Joel-Dahne:
I have not added tests for the ref constructors, it's a bit annoying to do since you cant construct them explicitly.

Joel-Dahne:
I can also add `set!` for relevant types here if we want to do that.

Joel-Dahne:
Okay, here is a pretty big rewrite of my started cleanup. I realised that we added more and more specialised setters, for example `set!(res::ArbLike, x::Rational; prec::Integer = precision(res))`, and that there was a lot of duplication with the code for the constructors. Now I have rewritten it so that I define extra setters in `setters.jl` and most constructors just call these directly. 

This makes `constructors.jl` a lot cleaner and separates the different parts of the code nicely I think. I adds a lot more constructors than we had before and all of them are now backed by a `set!` method as well.

src/constructors.jl
```diff
@@ -58,130 +19,45 @@ function Arb(str::AbstractString; prec::Integer = DEFAULT_PRECISION[])
...
-
-function Acb(::Irrational{:π}; prec::Integer = DEFAULT_PRECISION[])
-    res = Acb(prec = prec)
-    const_pi!(res)
-    return res
-end
+Base.zeros(x::T, n::Integer) where {T<:Union{Mag,Arf,Arb,Acb}} = [zero(x) for _ = 1:n]
+Base.ones(x::T, n::Integer) where {T<:Union{Mag,Arf,Arb,Acb}} = [one(x) for _ = 1:n]
+Base.zeros(::Type{T}, n::Integer) where {T<:Union{Mag,Arf,Arb,Acb}} = [zero(T) for _ = 1:n]
+Base.ones(::Type{T}, n::Integer) where {T<:Union{Mag,Arf,Arb,Acb}} = [one(T) for _ = 1:n]
```
kalmarek, line 63: This is a question: should we be returning ArbVectors/ArbMatrices here?

kalmarek (approved):

kalmarek:
@Joel-Dahne that's a really nice refactor here; initially I was internally ;) opposed to channeling everything through `set!`, but (especially with `midref` and `radref`) this really seems to be the right way. We'll need to have a comment on this in the docs though, since users will need to overload setters instead of constructors (?)

src/constructors.jl
Joel-Dahne: I don't think so. It would make the interface a bit confusing since what should it return in case you want more dimensions?  Writing this I realise that `zeros(Arb, n1, n2)` actually doesn't work... Neither does `zeros(Arb, n1, n2, n3, ...)`. This we should fix.

Joel-Dahne:
> @Joel-Dahne that's a really nice refactor here; initially I was internally ;) opposed to channeling everything through `set!`, but (especially with `midref` and `radref`) this really seems to be the right way. We'll need to have a comment on this in the docs though, since users will need to overload setters instead of constructors (?)

It's probably a good idea to mention it in the docs. Also so that they know that there are more `set!` than just those defined in Arb. I can add at least something about this now before merging.

README.md
```diff
@@ -110,6 +110,31 @@ becomes
 add!(z::ArbLike, x::ArbLike, y::Int; prec = precision(z))
 ```
 
+## Constructors and setters
+
+Arb defines a number of functions for setting something to a specific
+value, for example `void arb_set_si(arb_t y, slong x)`. All of these
+are renamed to `set!` and really on multiple dispatch to choose the
```
kalmarek, line 117: really → rely

README.md
```diff
@@ -110,6 +110,31 @@ becomes
...
+correct one. In addition to the ones defined in Arb there are a number
```
kalmarek, line 118: there is

README.md
```diff
@@ -110,6 +110,31 @@ becomes
...
+of ones added in Arblib to make it more convenient to work with. For
```
kalmarek, line 119: ones → methods of `set!`

PR #78, opened 2020-10-20
Allow to set Rational to Arb and Acb

saschatimme:
This uses the same strategy as for the constructors, i.e. performing a division by the denominator to get a result which is guaranteed to contain the rational number.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -60,8 +60,7 @@ end
 
 function Arb(x::Rational; prec::Integer = DEFAULT_PRECISION[])
     num = Arb(numerator(x); prec = prec)
-    denom = Arb(denominator(x); prec = prec)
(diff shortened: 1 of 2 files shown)


Joel-Dahne:
I thought about this when cleaning up the `constructors.jl`. Most of the constructors are just wrappers of `set!`, but some of them are not and in this case it might be nice to add a corresponding `set!` version. So far I think it's mainly `Rational` and several cases of `Complex`. In the future it might be more however, for example I have planned to add a constructor for `BigInt`. It could also be nice to add this for irrationals, so you can do `set!(x, π)` as well as `const_pi!(x)`. I could add all of these to my cleanup I started in that case.

Joel-Dahne (approved):

PR #79, opened 2020-10-22
Set version to 0.1

saschatimme:
We never had a 0.1 release l, so I guess we should start with this.

I would tag the release after this is merged. Ok? @kalmarek @Joel-Dahne

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "0.2.0"
+version = "0.1.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


kalmarek (approved):

kalmarek:
Fine with me!

Joel-Dahne:
I have still not understood exactly how releases work with Julia packages. But my understanding has been that the version is defined by the version number in `Project.toml` and not by any git-tags. So if it said `0.2.0` in the project file then the version was indeed `0.2.0`. Adding a git-tag with the release number is common practice but not required. So I believe we should make the release to be `0.3.0` and then also add a git-tag with that this time. I'm happy to be corrected in this though!

Joel-Dahne:
The terminology is indeed slightly confusing https://cdg.dev/tech/release-history/. But since we don't actually have any other users yet I'm happy to do whatever XD

kalmarek:
As long as we're not registered (or actually: don't have users:D) we're free to pick versions;

`0.1.0` is as good as any other and we should be fine as long as we (the only users of Arblib) didn't put any compat statements regarding `Arblib` in our `Project.toml`s

PR #80, opened 2020-10-22
Define arf_get_d and arf_get_si manually

saschatimme:
These two definitions are manually overridden in arbcall since they 
clash with our renaming rules. However, I need `arf_get_d`. So I 
manually defined the methods `get_d` resp. `get_si`. The idea is to keep 
the parth of original arb name which is necessary disambuige these two.


Joel-Dahne (approved):
I like the idea of keeping a set of manual overrides. Later on we could maybe handle this by allowing `arbcall` to take an alternative name as argument. For now this is fine!

PR #81, opened 2020-10-22
add generic constructors for Acb and Arb (through Arf)

kalmarek:
Add general constructors for `Real`, which go through `Arf`.

this allows you to write `Arb(big(1.))` and `Acb(big(1.0), big(2))`

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -64,6 +64,8 @@ function Arb(x::Rational; prec::Integer = DEFAULT_PRECISION[])
     return num
 end
 
+Arb(x::Real; prec::Integer = _precision(x)) = Arb(Arf(x, prec=prec))
+
 ## Acb
 for T in (acb_struct, Unsigned, Integer, Base.GMP.CdoubleMax)
     @eval begin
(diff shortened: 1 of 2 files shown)


saschatimme (approved):

Joel-Dahne (approved):
I think it's very reasonable to have a general constructor for `Acb` which goes through `Arb`, this will always give rigorous results. However and `Arb` constructors which goes through `Arf` could potentially lead to the result not enclosing the input. It's fine as long as all constructors for `Arf` give exact results, at the moment this is the case so it should be fine. But in the future we have to remember to enforce this.

I can note that I have done a lot of this work in the clean-up of the constructors that I did. But if you want to get this done before the release we can merge this and then I'll take it into account when finishing the clean-up.

kalmarek:
yeah, I'd like to have it in the released version, but this can also wait until your cleanup is merged (before we release, I suppose?)

Joel-Dahne:
I don't think I'll finish the cleanup in time for the release. I'm perfectly fine just taking care of them later.

PR #86, opened 2020-11-03
Enh/rand

kalmarek:
adds support for random api:
* everything goes through BigFloats
* radi are not randomized
* there is no rand for `Mag`s 

I also added `const ArbOrRef = Union{Arb, ArbRef}` but I won't insist on this change if there is some opposition to it ;)

There are couples of TODOs referencing `Random.SamplerBigFloat(precision(Arf))` which @rfourquet added/fixed very recently and probably will be in the next version. This  should simplify the code.

### DIFF
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -2,15 +2,13 @@ import Random
 
 # TODO: remove Random.CloseOpen01{BigFloat} parameters
 # TODO: remove setprecision
-# possibly fixed by
+# possibly fixed by https://github.com/JuliaLang/julia/pull/38169
 
 function Random.Sampler(
     RNG::Type{<:Random.AbstractRNG},
-    st::Random.SamplerType{TOrRef},
+    st::Random.SamplerType{T},
     n::Random.Repetition,
-) where {TOrRef<:Union{Arf,ArfRef,Arb,ArbRef,Acb,AcbRef}}
-
-    T = _nonreftype(TOrRef)
+) where {T<:Union{Arf,Arb,Acb}}
     return Random.SamplerSimple(
         Random.SamplerType{T}(),
         Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(precision(T)),
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -27,16 +27,17 @@ function Random.Sampler(
     )
 end
 
-Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{Random.SamplerType{Arf}}) =
     setprecision(BigFloat, sp.data.prec) do
         Arf(rand(rng, sp.data), prec = sp.data.prec)
     end
 
-Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arb}) where {A,B} =
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{Random.SamplerType{Arb}}) =
     setprecision(BigFloat, sp.data.prec) do
         Arb(rand(rng, sp.data), prec = sp.data.prec)
     end
-Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Acb}) where {A,B} =
+
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{Random.SamplerType{Acb}}) =
     setprecision(BigFloat, sp.data.prec) do
         Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
     end
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -1,9 +1,5 @@
-import Random
 
-const MagOrRef = Union{Mag,MagRef}
-const ArfOrRef = Union{Arf,ArfRef}
-const ArbOrRef = Union{Arb,ArbRef}
-const AcbOrRef = Union{Acb,AcbRef}
+import Random
 
 for (T, RefT) in ((Arf, ArfOrRef), (Arb, ArbOrRef), (Acb, AcbOrRef))
     @eval begin
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -1,22 +1,19 @@
 import Random
 
-for (T, RefT) in ((Arf, ArfOrRef), (Arb, ArbOrRef), (Acb, AcbOrRef))
-    @eval begin
-        function Random.Sampler(
-            RNG::Type{<:Random.AbstractRNG},
-            st::Random.SamplerType{<:$RefT},
-            n::Random.Repetition,
-        )
-            # return Random.SamplerSimple(
-            #     Random.SamplerType{Arf}(),
-            #     Random.SamplerBigFloat(precision(Arf)),
-            # )
-            res = setprecision(BigFloat, precision($T)) do
-                Random.SamplerSimple(Random.SamplerType{$T}(), Random.Sampler(RNG, BigFloat, n))
-            end
-            return res
-        end
+function Random.Sampler(
+    RNG::Type{<:Random.AbstractRNG},
+    st::Random.SamplerType{TOrRef},
+    n::Random.Repetition,
+) where {TOrRef<:Union{Arf,ArfRef,Arb,ArbRef,Acb,AcbRef}}
+    T = _nonreftype(TOrRef)
+    # return Random.SamplerSimple(
+    #     Random.SamplerType{T}(),
+    #     Random.SamplerBigFloat(precision(T)),
+    # )
+    res = setprecision(BigFloat, precision(T)) do
+        Random.SamplerSimple(Random.SamplerType{T}(), Random.Sampler(RNG, BigFloat, n))
     end
+    return res
 end
 
 Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -1,0 +1,65 @@
+import Random
+
+const MagOrRef = Union{Mag,MagRef}
+const ArfOrRef = Union{Arf,ArfRef}
+const ArbOrRef = Union{Arb,ArbRef}
+const AcbOrRef = Union{Acb,AcbRef}
+
+for (T, RefT) in ((Arf, ArfOrRef), (Arb, ArbOrRef), (Acb, AcbOrRef))
+    @eval begin
+        function Random.Sampler(
+            RNG::Type{<:Random.AbstractRNG},
+            st::Random.SamplerType{<:$RefT},
+            n::Random.Repetition,
+        )
+            #     TODO: SamplerBigFloat(precision(Arf))
+            #     return Random.SamplerSimple(
+            #         Random.SamplerType{Arf}(),
+            #         Random.SamplerBigFloat(precision(Arf))
+            #     )
+            res = setprecision(BigFloat, precision($T)) do
+                Random.SamplerSimple(Random.SamplerType{$T}(), Random.Sampler(RNG, BigFloat, n))
+            end
+            return res
+        end
+    end
+end
+
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
+    Arf(rand(rng, sp.data), prec = sp.data.prec)
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arb}) where {A,B} =
+    Arb(rand(rng, sp.data), prec = sp.data.prec)
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Acb}) where {A,B} =
+    Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
+
+function Random.rand(
+    rng::Random.AbstractRNG,
+    st::Random.SamplerTrivial{A,A},
+) where {A<:ArfOrRef}
+    # TODO: Arf(Random.SamplerBigFloat(precision(st[])))
+    x = setprecision(BigFloat, precision(st[])) do
+        rand(rng, BigFloat)
+    end
+    return Arf(x, prec = precision(st[]))
+end
+
+function Random.rand(
+    rng::Random.AbstractRNG,
+    st::Random.SamplerTrivial{A,A},
+) where {A<:ArbOrRef}
+    x = setprecision(BigFloat, precision(st[])) do
+        rand(rng, BigFloat)
+    end
+    # randomize radius?
+    return Arb(x, prec = precision(st[]))
+end
+
+function Random.rand(
+    rng::Random.AbstractRNG,
+    st::Random.SamplerTrivial{A,A},
+) where {A<:AcbOrRef}
+    re, im = setprecision(BigFloat, precision(st[])) do
+        rand(rng, BigFloat, 2) # TODO: Random.SamplerBigFloat(precision(st[]))
+    end # TODO: Random.SamplerBigFloat(precision(st[]))
+    return Acb(re, im; prec = precision(st[]))
+end
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -1,58 +1,44 @@
 import Random
 
+# TODO: remove Random.CloseOpen01{BigFloat} parameters
+# TODO: remove setprecision
+# possibly fixed by
+
 function Random.Sampler(
     RNG::Type{<:Random.AbstractRNG},
     st::Random.SamplerType{TOrRef},
     n::Random.Repetition,
 ) where {TOrRef<:Union{Arf,ArfRef,Arb,ArbRef,Acb,AcbRef}}
+
     T = _nonreftype(TOrRef)
-    # return Random.SamplerSimple(
-    #     Random.SamplerType{T}(),
-    #     Random.SamplerBigFloat(precision(T)),
-    # )
-    res = setprecision(BigFloat, precision(T)) do
-        Random.SamplerSimple(Random.SamplerType{T}(), Random.Sampler(RNG, BigFloat, n))
-    end
-    return res
+    return Random.SamplerSimple(
+        Random.SamplerType{T}(),
+        Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(precision(T)),
+    )
+end
+
+function Random.Sampler(
+    RNG::Type{<:Random.AbstractRNG},
+    x::TOrRef,
+    n::Random.Repetition,
+) where {TOrRef<:Union{Arf,ArfRef,Arb,ArbRef,Acb,AcbRef}}
+    T = _nonreftype(TOrRef)
+    return Random.SamplerSimple(
+        Random.SamplerType{T}(),
+        Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(precision(x)),
+    )
 end
 
 Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
-    Arf(rand(rng, sp.data), prec = sp.data.prec)
+    setprecision(BigFloat, sp.data.prec) do
+        Arf(rand(rng, sp.data), prec = sp.data.prec)
+    end
+
 Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arb}) where {A,B} =
-    Arb(rand(rng, sp.data), prec = sp.data.prec)
+    setprecision(BigFloat, sp.data.prec) do
+        Arb(rand(rng, sp.data), prec = sp.data.prec)
+    end
 Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Acb}) where {A,B} =
-    Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
-
-function Random.rand(
-    rng::Random.AbstractRNG,
-    st::Random.SamplerTrivial{A,A},
-) where {A<:ArfOrRef}
-    # TODO: Arf(rand(rng, Random.SamplerBigFloat(precision(st[]))))
-    x = setprecision(BigFloat, precision(st[])) do
-        rand(rng, BigFloat)
+    setprecision(BigFloat, sp.data.prec) do
+        Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
     end
-    return Arf(x, prec = precision(st[]))
-end
-
-function Random.rand(
-    rng::Random.AbstractRNG,
-    st::Random.SamplerTrivial{A,A},
-) where {A<:ArbOrRef}
-    # TODO: Arb(rand(rng, Random.SamplerBigFloat(precision(st[]))))
-    x = setprecision(BigFloat, precision(st[])) do
-        rand(rng, BigFloat)
-    end
-    # randomize radius?
-    return Arb(x, prec = precision(st[]))
-end
-
-function Random.rand(
-    rng::Random.AbstractRNG,
-    st::Random.SamplerTrivial{A,A},
-) where {A<:AcbOrRef}
-    # TODO: rand(rng, Random.SamplerBigFloat(precision(st[])), 2)
-    re, im = setprecision(BigFloat, precision(st[])) do
-        rand(rng, BigFloat, 2)
-    end
-    return Acb(re, im; prec = precision(st[]))
-end
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -1,4 +1,3 @@
-
 import Random
 
 for (T, RefT) in ((Arf, ArfOrRef), (Arb, ArbOrRef), (Acb, AcbOrRef))
@@ -8,11 +7,10 @@ for (T, RefT) in ((Arf, ArfOrRef), (Arb, ArbOrRef), (Acb, AcbOrRef))
             st::Random.SamplerType{<:$RefT},
             n::Random.Repetition,
         )
-            #     TODO: SamplerBigFloat(precision(Arf))
-            #     return Random.SamplerSimple(
-            #         Random.SamplerType{Arf}(),
-            #         Random.SamplerBigFloat(precision(Arf))
-            #     )
+            # return Random.SamplerSimple(
+            #     Random.SamplerType{Arf}(),
+            #     Random.SamplerBigFloat(precision(Arf)),
+            # )
             res = setprecision(BigFloat, precision($T)) do
                 Random.SamplerSimple(Random.SamplerType{$T}(), Random.Sampler(RNG, BigFloat, n))
             end
@@ -32,7 +30,7 @@ function Random.rand(
     rng::Random.AbstractRNG,
     st::Random.SamplerTrivial{A,A},
 ) where {A<:ArfOrRef}
-    # TODO: Arf(Random.SamplerBigFloat(precision(st[])))
+    # TODO: Arf(rand(rng, Random.SamplerBigFloat(precision(st[]))))
     x = setprecision(BigFloat, precision(st[])) do
         rand(rng, BigFloat)
     end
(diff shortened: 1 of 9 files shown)


kalmarek:
oh, and I played with Flint rand:
```julia
mutable struct _FlintRandState
    ptr::Ptr{Cvoid}

    function _FlintRandState()
        res = new(ccall(@libflint(flint_rand_alloc), Ptr{Cvoid}, ()))
        ccall(@libflint(flint_randinit), Cvoid, (Ptr{Cvoid},), res.ptr)
        finalizer(_flint_rand_free, res)
        res
    end
end

function _flint_rand_free(fr::_FlintRandState)
    ccall(@libflint(flint_rand_clear), Cvoid, (Ptr{Cvoid},), fr.ptr)
    ccall(@libflint(flint_rand_free), Cvoid, (Ptr{Cvoid},), fr.ptr)
end
```
e.g.
```julia
fr = Arblib._FlintRandomState()
y = Arf()
ccall(Arblib.@libarb(arf_randtest), Cvoid, (Ref{Arblib.arf_struct}, Ptr{Cvoid}, Cint, Cint), y, fr.ptr, 96, 25);
```
but the results are far from you'd expect from `rand` :)

Joel-Dahne:
Lets stay with the rand interface from Julia then! It's also nice that it respects the internal state of the Julia generator.

I really want something like `const ArbOrRef = Union{Arb, ArbRef}`. I'm not entirely sure about the naming but I have not been able to figure out anything better anyway. So I would say lets go with it.

src/rand.jl
```diff
@@ -0,0 +1,61 @@
...
+            #         Random.SamplerType{Arf}(),
+            #         Random.SamplerBigFloat(precision(Arf))
+            #     )
+            res = setprecision(BigFloat, precision($T)) do
+                Random.SamplerSimple(Random.SamplerType{$T}(), Random.Sampler(RNG, BigFloat, n))
+            end
+            return res
+        end
+    end
+end
```
Joel-Dahne, line 22: If we had a method that given a ref-type returned the base-type then this could be written without the `@eval` right? I think that would make it slightly cleaner. 

I have also wanted a method like that in some other cases, in general converting between the base and ref type. We don't necessarily have to add it now though.

src/rand.jl
```diff
@@ -0,0 +1,61 @@
...
+    return Arb(x, prec = precision(st[]))
+end
+
+function Random.rand(
+    rng::Random.AbstractRNG,
+    st::Random.SamplerTrivial{A,A},
+) where {A<:AcbOrRef}
+    re, im = setprecision(BigFloat, precision(st[])) do
+        rand(rng, BigFloat, 2) # TODO: Random.SamplerBigFloat(precision(st[]))
+    end # TODO: Random.SamplerBigFloat(precision(st[]))
```
Joel-Dahne, line 59: Maybe reformat these TODOs :)

Joel-Dahne (approved):

src/rand.jl
kalmarek: yep, I could write something like
```julia
_nonreftype(::Type{ArfOrRef}) = Arf
_nonreftype(::Type{ArbOrRef}) = Arb
_nonreftype(::Type{AcbOrRef}) = Acb

function Random.Sampler(
    RNG::Type{<:Random.AbstractRNG},
    st::Random.SamplerType{TOrRef},
    n::Random.Repetition,
) where TOrRef <: Union{Arf, ArfRef, Arb, ArbRef, Acb, AcbRef}
    T = _nonreftype(TOrRef)
    # return Random.SamplerSimple(
    #     Random.SamplerType{T}(),
    #     Random.SamplerBigFloat(precision(T)),
    # )
    res = setprecision(BigFloat, precision(T)) do
        Random.SamplerSimple(Random.SamplerType{T}(), Random.Sampler(RNG, BigFloat, n))
    end
    return res
end
```
Should I go for it here?

btw. In the longer run I really think we should separate our code generation into `Meta` module, so it doesn't pollute `Arblib` namespace and prefix non-Arb helper functions by `_`...

Joel-Dahne: I would say go for it here! 

I do agree that it's annoying when all the actual code and all the meta code is mixed. Trying to have internal methods with an underscore is a good way to at least be able to tell them apart quickly. All the different type-unions we have defined also makes some things hard to read.

src/rand.jl
```diff
@@ -0,0 +1,58 @@
...
+    Arf(rand(rng, sp.data), prec = sp.data.prec)
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arb}) where {A,B} =
+    Arb(rand(rng, sp.data), prec = sp.data.prec)
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Acb}) where {A,B} =
+    Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
+
+function Random.rand(
+    rng::Random.AbstractRNG,
+    st::Random.SamplerTrivial{A,A},
+) where {A<:ArfOrRef}
```
rfourquet, line 29: Maybe naive question, but why define `rand` for "scalars" here? Is it like calling `rand(big(1.3))`  and getting a random `BigFloat` of the same precision as the input?

src/rand.jl
```diff
@@ -0,0 +1,58 @@
...
+    # TODO: Arf(rand(rng, Random.SamplerBigFloat(precision(st[]))))
+    x = setprecision(BigFloat, precision(st[])) do
+        rand(rng, BigFloat)
```
rfourquet, line 32: I guess you are working around the bug in `Random` which doesn't create a `BigFloat` according to the precision set in the `Sampler`? But I guess you could still store the `SamplerBigFloat` as a `.data` field of a `SamplerSimple`, and do 
```
x = setprecision(BigFloat, sp.data.prec) do
        rand(rng, sp.data)
end
```
(but anyway, the random generation is quite more expensive that the `Sampler` creation`, so the benefit would not be huge).

src/rand.jl
```diff
@@ -0,0 +1,58 @@
...
+    #     Random.SamplerBigFloat(precision(T)),
+    # )
+    res = setprecision(BigFloat, precision(T)) do
+        Random.SamplerSimple(Random.SamplerType{T}(), Random.Sampler(RNG, BigFloat, n))
+    end
+    return res
+end
+
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
+    Arf(rand(rng, sp.data), prec = sp.data.prec)
```
rfourquet, line 20: Isn't it necessary here also to set the `BigFloat` precision according to `sp` before calling `rand` ?

src/rand.jl
kalmarek: yes, if we already have `x = Arb("1.1", prec=128)`, I'd like to have `precision(rand(x)) == precision(x)`

src/rand.jl
kalmarek: that's what TODO says, but at the moment I don't know how to easily create `SamplerBigFloat` with the given precision ;) any advice on this?

src/rand.jl
kalmarek: here `sp.data` is `Sampler(RNG, BigFloat, n)` which was created within `setprecision` `do` block`. I understood that this will fix the precision?

src/rand.jl
rfourquet: Ok; I just find it weird, as `rand(x)` doesn't depend on the value of `x`, just on its precision. But alternatives I can think of are obviously more verbose :shrug:

src/rand.jl
rfourquet: Yep:
```
setprecision(BigFloat, prec) do
    Random.Sampler(RNG, BigFloat)
end
```
or 
```
Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(prec)
```

src/rand.jl
rfourquet: Yes, this sets the precision of the `Sampler`, but until very recently, there is a bug such that `rand(sp)` will create a `BigFloat` with the current precision instead of the precision of the `sp` (so in particular this will crash sometimes when `sp.prec` is bigger than the current precision). I though you were refering to this bug in the TODOS. So this is fixed in 1.6, if you like I can push the fix to the 1.5 backport branch.

src/rand.jl
kalmarek: indeed, but since we're not carrying precision in the type, that was the easiest way I could thought how to pass it.
I think this is useful in iterative algorithms where you double precision of computations every step, and you need randomness that stretches far outside Float64 range

src/rand.jl
```diff
@@ -0,0 +1,48 @@
...
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arb}) where {A,B} =
+    Arb(rand(rng, sp.data), prec = sp.data.prec)
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Acb}) where {A,B} =
+    Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
+
+function Random.rand(
+    rng::Random.AbstractRNG,
+    st::Random.SamplerTrivial{A,A},
+) where {A<:ArfOrRef}
+    x = rand(rng, Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(precision(st[])))
```
rfourquet, line 29: This is equivalent as what you had before; what I meant is that you could store the `SamplerBigFloat` in a `SamplerSimple` like you do for the types version (`rand(Arb)`). Again, this won't be a huge improvement, but will manifest for example when creating arrays.

kalmarek: I see, you meant creating `SamplerSimple` instead of relying on the automatic creation of `SamplerTrivial`.

kalmarek:
this seems to work, but fails spectacularly for other than the default precisions:

```julia
julia> y = big(0.)
       for i in 1:1_000_000   
           x = rand(BigFloat, 2,2)
           y += x[1,1]
           y -= 0.5
       end
       y/1_000_000
0.000150236616259007470914305217465882434760984169073877357131989077811491794938099

julia> y = Arb(0)
       for i in 1:1_000_000   
           x = rand(Arb, 2,2)
           y[] += x[1,1]
           y[] -= 0.5
       end
       y/1_000_000
[-0.00010688314298852862159639023199183028037121222759491026767205766655184876 +/- 2.02e-75]

julia> y = Arb(0)
       for i in 1:1_000_000   
           x = rand(Arb(), 2,2)
           y[] += x[1,1]
           y[] -= 0.5
       end
       y/1_000_000
[0.00021218346835001610055588556821600451678368979176801618968402745035149904 +/- 2.80e-75]

julia> y = Arb(0)
       for i in 1:1_000_000   
           x = rand(Arb(prec=64), 2,2)
           y[] += x[1,1]
           y[] -= 0.5
       end
       y/1_000_000
[-0.39057934971145324865126392341311022509822032354151069212056277134594156 +/- 5.87e-72]

julia> y = Arb(0)
       for i in 1:1_000_000   
           x = rand(Arb(prec=512), 2,2)
           y[] += x[1,1]
           y[] -= 0.5
       end
       y/1_000_000

signal (11): Segmentation fault
in expression starting at REPL[24]:2
jl_gc_pool_alloc at /buildworker/worker/package_linux64/build/src/gc.c:1148
jl_gc_alloc_ at /buildworker/worker/package_linux64/build/src/julia_internal.h:277 [inlined]
jl_alloc_string at /buildworker/worker/package_linux64/build/src/array.c:504
_string_n at ./strings/string.jl:60 [inlined]
BigFloat#1 at ./mpfr.jl:116
BigFloat at ./mpfr.jl:112 [inlined]
_rand at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/generation.jl:60 [inlined]
_rand at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/generation.jl:80
rand at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/generation.jl:98 [inlined]
rand at /home/kalmar/.julia/dev/Arblib/src/rand.jl:32 [inlined]
rand! at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/Random.jl:271
rand! at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/Random.jl:266 [inlined]
rand at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/Random.jl:279
rand at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/Random.jl:280 [inlined]
rand at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/Random.jl:283 [inlined]
top-level scope at ./REPL[24]:3
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:834
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:790
jl_toplevel_eval_in at /buildworker/worker/package_linux64/build/src/toplevel.c:883
eval at ./boot.jl:331
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2214 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
eval_user_input at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:134
repl_backend_loop at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:195
start_repl_backend at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:180
#run_repl#37 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:292
run_repl at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2231 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
#807 at ./client.jl:399
jfptr_YY.807_64838.clone_1 at /opt/julia-1.5.2/lib/julia/sys.so (unknown line)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2214 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1690 [inlined]
do_apply at /buildworker/worker/package_linux64/build/src/builtins.c:655
jl_f__apply_latest at /buildworker/worker/package_linux64/build/src/builtins.c:705
#invokelatest#1 at ./essentials.jl:710 [inlined]
invokelatest at ./essentials.jl:709 [inlined]
run_main_repl at ./client.jl:383
exec_options at ./client.jl:313
_start at ./client.jl:506
jfptr__start_60376.clone_1 at /opt/julia-1.5.2/lib/julia/sys.so (unknown line)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2214 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
unknown function (ip: 0x401931)
unknown function (ip: 0x401533)
__libc_start_main at /usr/lib/libc.so.6 (unknown line)
unknown function (ip: 0x4015d4)
Allocations: 505967705 (Pool: 505964740; Big: 2965); GC: 810
[1]    221150 segmentation fault (core dumped)  julia --project
```

also note that the result for `rand(Arb(prec=64))` is much inflated....
@Joel-Dahne is there something we should look for when setting Arbs from mpfrs?

kalmarek:
and when not crashing, you can clearly see that something is very wrong with those numbers:
```julia
julia> using Arblib

julia> rand(Arb(2, prec=64), 5,5)
5×5 Array{Arb,2}:
  [0.7500000000220552465 +/- 1.80e-20]  [-0.4999999999999999998 +/- 3.74e-20]  …  [0.8009728521756187547 +/- 1.76e-22]
 [-0.4999999999999999998 +/- 2.59e-20]  [-0.4999999999999999998 +/- 3.74e-20]     [0.8749999999672581907 +/- 3.62e-20]
  [0.7500000000220552465 +/- 1.80e-20]  [-0.5000000000000000000 +/- 3e-24]        [0.8984375000000000000 +/- 8.50e-22]
  [0.5240253969810737544 +/- 3.22e-20]   [0.7414182195258744657 +/- 1.83e-20]     [0.7764110785818262433 +/- 1.26e-20]
 [-0.4999999999999999998 +/- 1.03e-20]  [-0.2253371962693598937 +/- 1.13e-20]     [0.7500018953262497234 +/- 3.92e-20]

julia> rand(Arb(2, prec=64), 5,5)
5×5 Array{Arb,2}:
 [-0.4999924185912108911 +/- 5.91e-21]  [-0.4999924185912108911 +/- 5.91e-21]  …  [-0.4999924185912108911 +/- 5.91e-21]
 [-0.4999924185912108911 +/- 5.91e-21]   [0.9937104128060809671 +/- 2.97e-20]      [0.9937104128060809671 +/- 2.97e-20]
 [-0.4999924185912108911 +/- 5.91e-21]  [-0.4999924185912108911 +/- 5.91e-21]     [-0.4999924185912108911 +/- 5.91e-21]
  [0.9937104128060809671 +/- 2.97e-20]   [0.9937104128060809671 +/- 2.97e-20]      [0.9937104128060809671 +/- 2.97e-20]
  [0.9937104128060809671 +/- 2.97e-20]   [0.9937104128060809671 +/- 2.97e-20]      [0.9937104128060809671 +/- 2.97e-20]

julia> rand(Arb(2, prec=64), 5,5)
5×5 Array{Arb,2}:
 [-0.4999924185941800455 +/- 4.22e-20]   [0.9937104128060809671 +/- 2.97e-20]  …   [0.9937104128060809671 +/- 2.97e-20]
 [-0.5000000000000000000 +/- 3e-24]      [0.9937104108419134719 +/- 3.29e-20]     [-0.4999924185912108911 +/- 5.91e-21]
 [-0.4999924185912108911 +/- 5.91e-21]  [-0.4999924185912108911 +/- 5.91e-21]     [-0.4999924185912108911 +/- 5.91e-21]
 [-0.4999924185912108911 +/- 5.91e-21]  [-0.4999924185912108911 +/- 5.91e-21]     [-0.4999924185912108911 +/- 5.91e-21]
  [0.9937104128060809671 +/- 2.97e-20]   [0.9937104128060809671 +/- 2.97e-20]     [-0.4999924185912108911 +/- 5.91e-21]

julia> rand(Arb(2, prec=64), 5,5)
5×5 Array{Arb,2}:

signal (11): Segmentation fault
in expression starting at none:0
_arf_set_round_mpn at /workspace/srcdir/arb-2.18.1/arf/set_round_mpn.c:52
arf_mul_rnd_down at /workspace/srcdir/arb-2.18.1/arf/mul_rnd_down.c:213
arb_mul_arf at /workspace/srcdir/arb-2.18.1/arb/mul.c:22
arb_mul at /workspace/srcdir/arb-2.18.1/arb/mul.c:71
arb_get_fmpz_mid_rad_10exp at /workspace/srcdir/arb-2.18.1/arb/get_fmpz_mid_rad_10exp.c:83
arb_get_str_parts at /workspace/srcdir/arb-2.18.1/arb/get_str.c:335
arb_get_str at /workspace/srcdir/arb-2.18.1/arb/get_str.c:453
arb_fprintn at /workspace/srcdir/arb-2.18.1/arb/fprint.c:34
printn at /home/kalmar/.julia/dev/Arblib/src/arbcall.jl:295 [inlined]
string_nice at /home/kalmar/.julia/dev/Arblib/src/show.jl:92
string_nice at /home/kalmar/.julia/dev/Arblib/src/show.jl:87 [inlined]
string_nice at /home/kalmar/.julia/dev/Arblib/src/show.jl:87
show at /home/kalmar/.julia/dev/Arblib/src/show.jl:4
unknown function (ip: 0x7f31daef429c)
```

Joel-Dahne:
I think it's an issue with `BigFloat` and not `Arblib.jl` at least. Adding a `@show x` in `set!(res::ArbLike, x::Union{MagLike,BigFloat})` makes it crash there instead with
```
julia> rand(Arb(2, prec=64))
get_str.c:157: MPFR assertion failed: size_s1 >= m
```
I have not understood how the `Random` framework works yet so I would have a hard time debugging this.

rfourquet:
I didn't look in details, but if this can indeed be a crash related `rand` here, then I would suggest trying my suggestion of setting `BigFloat` precision before calling `rand`. I will add a comment there.

src/rand.jl
rfourquet:
```suggestion
    setprecision(BigFloat, sp.data.prec) do 
        Arf(rand(rng, sp.data), prec = sp.data.prec)
    end
```

src/rand.jl
```diff
@@ -0,0 +1,35 @@
...
+    return Random.SamplerSimple(
+        Random.SamplerType{T}(),
+        Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(precision(x))
+    )
+end
+
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
+    Arf(rand(rng, sp.data), prec = sp.data.prec)
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arb}) where {A,B} =
+    Arb(rand(rng, sp.data), prec = sp.data.prec)
```
rfourquet, line 33:
```suggestion
    setprecision(BigFloat, sp.data.prec) do 
        Arb(rand(rng, sp.data), prec = sp.data.prec)
    end
```

src/rand.jl
```diff
@@ -0,0 +1,35 @@
...
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Acb}) where {A,B} =
+    Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
```
rfourquet, line 35:
```suggestion
    setprecision(BigFloat, sp.data.prec) do 
        Acb(rand(rng, sp.data), rand(rng, sp.data), prec = sp.data.prec)
    end
```

kalmarek:
> I didn't look in details, but if this can indeed be a crash related `rand` here, then it could be worth trying my suggestion of setting `BigFloat` precision before calling `rand`. I will add a comment there.

@rfourquet thanks, this seems to fix the issue! I was reluctant to do so initially, as `setprecision` is not thread safe and uses global variable, but if that's what it takes to fix it, then let it be ;)

rfourquet:
> this seems to fix the issue! I was reluctant to do so initially, as setprecision is not thread safe and uses global variable, but if that's what it takes to fix it, then let it be ;)

Cool, and you're welcome. If you really want to avoid using setprecision, I'm sure we can come up with an alternate, more involved, solution. And in julia 1.6 this will be fixed anyway.

Also, the nightly failures are real: `rand(MyType)` must return an object of type `MyType`, and in 1.6 a typeassert was added to help inference (which causes the build failure), but this just enforces what was already required by the API.

kalmarek:
@rfourquet no need to break your neckl;), we will most probably drop support < 1.6 once it's announced stable LTS ;)

I don't quite understand failures on nightly though:
```julia
julia> all(typeof(rand(T)) == T for T in (Arf, Arb, Acb))
true
```
(at least this holds on 1.5.2)

rfourquet:
> I don't quite understand failures on nightly though:

I guessed it's for `ArbRef` and friends, via the line: `T = _nonreftype(TOrRef)`

For this, you would need a dummy type of which you pass an instance. Using `RandomExtensions` for example, it could look like `rand(make(ArbRef))`.

kalmarek:
of course, silly me ;) @Joel-Dahne @saschatimme what do you think: RandomExtensions, or not allowing to randomize refs?

Joel-Dahne:
A think excluding ref types from random generation is fine for now. This is related to a bigger issue about how to treat ref types in general. For example how to treat `convert(ArbRef, x)`.

src/types.jl
```diff
@@ -248,12 +248,18 @@ for prefix in [:mag, :arf, :arb, :acb]
         Base.convert(::Type{Ptr{$TStruct}}, x::$TRef) = cstruct(x)
         Base.cconvert(::Type{Ref{$TStruct}}, x::$TRef) = cstruct(x)
 
+        _nonreftype(::Type{<:Union{$T,$TRef}}) = $T
+
```
Joel-Dahne, line 252: Assuming my comment about removing ref types above is accurate I guess this is no longer used? Could possibly keep it anyway for future use?

test/rand.jl
```diff
@@ -0,0 +1,18 @@
+@testset "Random" begin
+    @testset "rand $T" for (T, RefT) in ((Arf, ArfRef), (Arb, ArbRef), (Acb, AcbRef))
```
Joel-Dahne, line 2: You can remove the `RefT` here.

src/rand.jl
```diff
@@ -0,0 +1,42 @@
...
+    RNG::Type{<:Random.AbstractRNG},
+    x::TOrRef,
+    n::Random.Repetition,
+) where {TOrRef<:Union{Arf,ArfRef,Arb,ArbRef,Acb,AcbRef}}
+    T = _nonreftype(TOrRef)
+    return Random.SamplerSimple(
+        Random.SamplerType{T}(),
+        Random.SamplerBigFloat{Random.CloseOpen01{BigFloat}}(precision(x)),
+    )
+end
```
Joel-Dahne, line 28: Should we have the ref types here? I didn't quite follow the discussion about why to remove then so I could be wrong.

Joel-Dahne (changes requested):

src/types.jl
kalmarek: yes, this is not used in the code; I'll remove it

src/rand.jl
kalmarek: yes, I believe so:
* the first sampler is used when calling `rand(Arb, ...)`, i.e. **on type**: rand(::Type{T}); here julia-master typeasserts this call always to return object of type `T`, so we had conflict with `rand(ArbRef, ...)` returning honest `Arb`s.
* the second sampler is applicable when calling rand on value `rand(Arb(prec=64))`, but also `x = Acb(prec=64); rand(Arblib.realref(x),...)`

@rfourquet suggested this is not the most natural call, since value of `x` has nothing to do with the return value (it's there only for passing its precision), but it is what it is ;) @Joel-Dahne If you have a better idea how to handle this, let me know!

and yes, someone might follow interpret that since `Arb`s are intervals, `rand(Arb("1.1"))` should return an `Arf` in proximity of `1.1`. This would be against what `IntervalArithmetic` does though:
https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/cb987732830339a160c12c6d6af0ee2c0f0dbf21/src/rand.jl#L3

----
Another option would be to create `SamplerArb` cheap objects e.g. through `make` from `RandomExtensions` which then would carry the precision, but I think it's more natural to use value here.

Joel-Dahne: Oh I see, so it's still possible to use instances of `ArbRef` in the generation. This seems reasonable!

About this maybe not being the most natural call, since value of x has nothing to do with the return value (it's there only for passing its precision), I don't fully agree. It's in line with how `zero` and `one` works for example, though `zero` and `one` doesn't preserve the precision for `BigFloat`.

kalmarek:
I rebased on master and this is good to go

Joel-Dahne (approved):
Excellent!

src/rand.jl
```diff
@@ -0,0 +1,42 @@
...
+
+Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{A,B,Arf}) where {A,B} =
```
rfourquet, line 30: Sorry for not spotting that earlier, but you might want to constrain the `A` type here. For now it probably won't matter, but while Arblib "owns" `Arf`, it doesn't own `SamplerSimple` and doesn't have the exclusivity on using `Arf`. It means that I could depend on Arblib, and write a `Sampler` method which also generates `Arf`, and could therefore write exactly the same method, without realizing I'm overwriting this one (or write another one which introduces ambiguitites). So it's best with `SamplerSimple` to constrain the first type parameter:
```suggestion
Random.rand(rng::Random.AbstractRNG, sp::Random.SamplerSimple{Random.SamplerType{Arf},B,Arf}) where {B} =
```
Maybe keeping the trailing `{B,Arf}` as type parameters of `SamplerSimpler` also becomes unnecessary then.

kalmarek: thanks for the remark, I changed the methods appropriately

Issue #87, opened 2020-11-05
Printing in IJulia deadlocks


saschatimme:
On Julia 1.5.2 in IJulia 1.22 (latest version) 

```julia
show(Acb(2.3));
```
deadlocks. In the terminal everything is fine. Anybody any clue what is going on here?

PR #88, opened 2020-11-05
Fix show+constructors

saschatimme:
This fixes the very annoying #87.
There seems no way to make our current `show` strategy work in IJulia.
The strategy is fundamentally flawed for IJulia since we don't have a guarantee that the C-std is actually flushed when we end the stdout redirect (e.g. https://github.com/JuliaLang/IJulia.jl/issues/238).

Thus I used the `arb_get_str` method for `Arb`, convert to BigFloat for `Arf` and, if we are in IJulia, to Float64 for `Mag`.
For `Acb` we need to define the printing manually. Here I ommited the imaginary part if it is zero which is conistent with Arb but not Julia.

I also added a `BigFloat(::Arf)` since I needed it and moved the `Float64/BigFloat` constructors to `constructors.jl` where they seemed to belong better.

I hope we can merge this quickly and get a new release out since my new certification code in HomotopyContinuation.jl is barely usable in IJulia right now (in the sense that if you print the wrong thing your session is over).

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -50,7 +50,8 @@ function Acb(re::AbstractString, im::AbstractString; prec::Integer = DEFAULT_PRE
     return res
 end
 
-Base.Int(x::ArfLike; rnd::Union{arb_rnd,RoundingMode} = RoundNearest) = get_si(x, rnd)
+Base.Int(x::ArfLike; rnd::Union{arb_rnd,RoundingMode} = RoundNearest) =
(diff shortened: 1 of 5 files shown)


kalmarek (changes requested):

saschatimme:
@kalmarek thanks for the review

saschatimme:
I fixed another bug and corrected some tests. I think this should now be good to go.

kalmarek (approved):

PR #91, opened 2020-11-09
Fix set rational

saschatimme:
#76 broke the assignment of Rationals to Vectors/Matrices.
I think previously the computation was also not fully correct in the sense that the conversion of `Rational` to `Arb` used always the global precision and not the precision of the target.
I changed the `setindex!` methods to always create first an `ArbRef` resp. `AcbRef` and to perform `set!` on these. By this way we carry around the precision of the target.

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -46,7 +46,6 @@ Base.@propagate_inbounds function Base.setindex!(
     j::Integer,
 )
     @boundscheck checkbounds(A, i, j)
-    # A.arb_mat[i, j] = x
     ref(A, i, j)[] = x
     return x
 end
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -46,7 +46,8 @@ Base.@propagate_inbounds function Base.setindex!(
     j::Integer,
 )
     @boundscheck checkbounds(A, i, j)
-    A.arb_mat[i, j] = x
+    # A.arb_mat[i, j] = x
+    ref(A, i, j)[] = x
     return x
 end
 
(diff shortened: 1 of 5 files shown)


saschatimme:
Since Arb `0.1.1` broke the certification in `HomotopyContinuation.jl` I would also appreciate a quick bugfix release.

src/matrix.jl
```diff
@@ -46,7 +46,8 @@ Base.@propagate_inbounds function Base.setindex!(
     j::Integer,
 )
     @boundscheck checkbounds(A, i, j)
-    A.arb_mat[i, j] = x
+    # A.arb_mat[i, j] = x
```
Joel-Dahne, line 49: Could remove this comment.

kalmarek, line 49:
```suggestion
```

Joel-Dahne (approved):

kalmarek (changes requested):

Joel-Dahne:
I realised another problem with the interface. It currently returns `x` and not `A[i, j]` as you would expect. For example `M[1, 1] = 1//3` returns `1//3` and not `Arb(1//3)` as expected.

saschatimme:
But I think this is consistent with Julia
```
julia> A = zeros(ComplexF64, 2, 2)

A2×2 Array{Complex{Float64},2}:
 0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im

julia> 

julia> A[1,1] = 2
2

```

kalmarek (approved):

PR #97, opened 2020-11-18
Fix/integration

Joel-Dahne:
This fixes a typo in the integration and adds a missing broken test.

### DIFF
--- a/src/calc_integrate.jl
+++ b/src/calc_integrate.jl
@@ -181,7 +181,7 @@ function integrate!(
     opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
 )
     # rtol ≈ 2^(-rel_goal)
-    rel_goal = iszero(rtol) ? prec : max(-floor(Int, log2(atol)), 0)
+    rel_goal = iszero(rtol) ? prec : max(-floor(Int, log2(rtol)), 0)
 
     if !check_analytic && !take_prec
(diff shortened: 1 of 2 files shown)


src/calc_integrate.jl
```diff
@@ -181,7 +181,7 @@ function integrate!(
     opts::Union{Ptr{Cvoid},calc_integrate_opt_struct} = C_NULL,
 )
     # rtol ≈ 2^(-rel_goal)
-    rel_goal = iszero(rtol) ? prec : max(-floor(Int, log2(atol)), 0)
+    rel_goal = iszero(rtol) ? prec : max(-floor(Int, log2(rtol)), 0)
```
Joel-Dahne, line 184: Another thing I notices is that since we use `log2` when computing the relative tolerance it wont work if `rtol` is any of the `Arblib.jl`types since we haven't defined that. Since we don't care to much about the exact value we could maybe just rewrite it as `log(rtol)*1.4426950408889634`? Here `1.4426950408889634 = 1/log(2)`.

kalmarek: very good! that was written hastily just to get something going and I never looked back ;)

Joel-Dahne: I changed my mind. Neither `log` nor `floor` is defined for `Mag` or `Arf` at the moment anyway, so the gain of switching away from `log2` wouldn't be very high. Then I think it's better to keep it like it is for clarity.

kalmarek (approved):

PR #98, opened 2020-11-21
Add parsing of acb_elliptic and acb_dirichlet

Joel-Dahne:
Apart from adding `acb_elliptic.h` and `acb_dirichlet.h` this also leads to a few other minor changes to the parsed functions. The reason is that I'm parsing against a newer (the latest) version of Arb. In general I'm not sure how we want to handle this. At the moment we can't parse against the version used by `Arb_jll` because the documentation has been fixed since then. However now we for example parse `csc_pi` which doesn't exist in the version `Arb_jll` uses. 

I realised that the `Arb_jll` version is quite old, do you know how it works to update that?


kalmarek:
At the moment I'd leave those `arbcall`s as they are, the chances that someone will stumble upon these is really slim (alternatively we may open an issue here as a reminder). A possible solution would be to "fork" arblib docs at `2.18.1` and pile our fixes there, to be reversed/overwritten later, but since you've been maintaining this part, I'm happy with any solution that works for us ;)

To me it seems that `Arblib_jll` is up to date with `Arblib`: both on version `2.18.1`? Probably to have things straighten out we need to wait for the next release of `Arblib` containing (doc)fixes so far; I hope it's ok to ping @fredrik-johansson for a `2.18.2` (`2.18.1` is from half a year ago).

fredrik-johansson:
If it's OK to wait a few days, I think I will make a 2.19.0 release soon.

Joel-Dahne:
Excellent! If the 2.19.0 release is soon then there are no issues at all. 

But I agree, the chance that someone would decide to use exactly `csc_pi`  seem very slim ;)

I missed that `Arblib_jll` was actually up to date with Arb. I usually follow the master branch of Arb and didn't realise there hadn't been a release for some time. Then everything should be fine once 2.19.0 is released.

kalmarek (approved):

PR #99, opened 2020-11-28
New constructors for intervals

Joel-Dahne:
I have wanted to add some type of constructor for intervals. After thinking about it for a while I believe having tuples correspond to intervals would make the most sense. With this change you can construct the ball containing `a` and `b` with `Arb((1, 2))`. For example
```julia
julia> Arb((1, 2))
[+/- 2.01]

julia> Arb((1, π))
[+/- 3.15]

julia> Arb((3.141592, π))
[3.141592 +/- 6.54e-7]

julia> Arb((ℯ, π))
[3e+0 +/- 0.282]
```
As for the other constructors it uses a `set!` method, `Arblib.set!(Arb(), (5, 6))` for example. No matter what types you give in it the resulting ball should always include both the inputs. It works for `Acb` as well
```julia
julia> Acb((1, 2))
[+/- 2.01]

julia> Acb((1, 2), (float(π), π))
[+/- 2.01] + [3.1415926535897932 +/- 8.41e-17]im

julia> Acb((Acb(1, 1), Acb(2, 2)))
[+/- 2.01] + [+/- 2.01]im
```
I have yet to add tests for the constructors (the setters have tests) and information about it to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,9 @@ are renamed to `set!` and rely on multiple dispatch to choose the
 correct one. In addition to the ones defined in Arb there is a number
 of methods of `set!` added in Arblib to make it more convenient to
 work with. For example there are setters for `Rational` and all
-irrationals defined in `Base.MathConstants`.
+irrationals defined in `Base.MathConstants`. For `Arb` there is also a
+setter which takes a tuple `(a, b)` representing an interval and
+returns a ball containing this interval.
 
 Almost all of the constructors are simple wrappers around these
 setters. This means that it's usually more informative to look at the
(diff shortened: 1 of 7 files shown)


kalmarek (approved):
I'm not swayed by `set!(res::AcbLike, (a, b)::NTuple{2,AcbLike})`. The only reasonable (mathematical) meaning is `union!` and it's hard for me to imagine that somebody will naturally turn into this constructor instead of `union!`.

On the other hand it's probably nice to have from the point of view of generic functions that will accept both `Arb`s and `Acb`s, so I'm not going to protest either ;)

Joel-Dahne:
You have a good point! I have removed the constructors of `Acb` from a pair of complex values. Added a few more tests and a small comment in the README. Think it should be good to go now!

PR #100, opened 2020-12-03
replaced `Clong`, `Culong` to match Julia

JeffreySarnoff:
Julia uses `Cint`, `Cuint` for 32bit Integers and `Clonglong`, `Culonglong` for 64bit integers.  I replaced `Clong` and `Culong` accordingly.  See "system independent types" in https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#mapping-c-types-to-julia.  The way it had been written generated this on v1.5.3 (Windows, 64bit)
```
julia> using Arblib
[ Info: Precompiling Arblib [fb37089c-8514-4489-9461-98f9c8763369]
WARNING: Method definition jltype(Arblib.Carg{Int32}) in module Arblib at C:\Users\jas\.julia\packages\Arblib\GWaO3\src\arbcall.jl:103 overwritten at C:\Users\jas\.julia\packages\Arblib\GWaO3\src\arbcall.jl:104.
  ** incremental compilation may be fatally broken for this module **
```


JeffreySarnoff:
I had not realized you were using `Clong` in many places.
I need to update each occurance.  Did you intend 32bits or 64bits in these uses?

Joel-Dahne:
Thank you very much for taking a look at this! We never tried running it on Windows so it's good that someone can do that!

I believe that every time we use `Clong` we are referring to what in Arb is called `slong`. At the moment we make use of `Clong`, `Cuint` and `Cint`, I find no occurance of `Cuint`. So from what I understand we have to replace all occurances of `Clong` with `Clonglong` and `Culong` with `Culonglong`? On my system (Ubuntu) they are indeed aliased to the same thing, but this is not the case on Windows then I guess.

kalmarek:
this whole business of different Ints is a general mess :D (CxxWrap solved it by adding its own CxxLong with I'd like not to repeat;))

@JeffreySarnoff how does `Clonglong` work on `i686` arches where `Arblib`s `slong` is `32`-bit integer?
but anyway: I think we only care about 64bit oses, so changing those types could be done without much repercussions.

Or we could define those using 
```julia
@static if Sys.iswindows()
    ...
else
   ....
end
```
?

Joel-Dahne:
If we cover 64bit Window, Mac and Linux I think we can be mostly satisfied. If we can avoid a `isWindows()` check that would be nicer I think, but we do what we have to do.

saschatimme:
I think we should first extend our CI to a 32 bit configuration to see whether something is actually broken.

Joel-Dahne:
I looked a bit but couldn't find anything about using a 32-bit OS in the GitHub actions. I could add windows to our tests and make the proposed switch to `Clonglong` to see if that works at least. Is it okay if I hijack this PR for it @JeffreySarnoff?

JeffreySarnoff:
Hijack away

Joel-Dahne:
I rebased on master and added Windows to the CI, as expected those tests failed at first. I did a search and replace for `Clong` to `Clonglong` as well as `Culong` to `Culonglong` and as you can see all tests passes now. However I don't think it will wok on 32-bit systems anymore, since then `Clonglong` is 64-bit but Arb expects 32-bits. 

As you can read in the Arb documentation, https://www.arblib.org/issues.html#integer-types, it uses `slong` and `ulong` which are always 32-bits on 32-bit systems and 64-bits on 64-bit systems. Would one solution be to use `Int` directly? From the Julia documentation, https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/, that is supposed to be 32-bits on 32-bit systems and 64-bits on 64-bit systems as well. It makes it less clear that the type is supposed to interact with C, but might make it work everywhere. Any thoughts?

saschatimme:
I think it makes indeed sense to use the Julia `Int` and `UInt` types since these have the exact same guarantees as the integer types used in Arb.

JeffreySarnoff:
Yes. That works.

Joel-Dahne:
Are you able to test this a bit locally as well @JeffreySarnoff ? I don't have access to a Windows system so I have only the tests in GitHub to rely on. There is no rush for it though!

JeffreySarnoff:
I will test it and let you know how it goes.

saschatimme (approved):

saschatimme:
I think we can merge this since windows is currently broken anyways, so this is clearly progress :)

PR #102, opened 2020-12-04
add expm1

kalmarek:


Joel-Dahne (approved):

kalmarek:
I bumped to `0.2.1` so that we could tag (version numbers are free ;)

PR #107, opened 2020-12-18
Make compat entries tighter

katatimme:
After a second thought, we probably want to have the tightest compat entries for now.


saschatimme (approved):

kalmarek:
@katatimme @saschatimme  it's great to have a new contributor returning :)

Issue #94, opened 2020-11-15
rand allocations


kalmarek:
as observed in #82:
```julia
julia> @btime ArbMatrix(rand(Arb(prec=256), 100,100));
  4.085 ms (50007 allocations: 3.21 MiB)

julia> @btime ArbMatrix(rand(Arb, 100,100));
  4.507 ms (60005 allocations: 4.27 MiB)
```
i.e. 6 allocations per entry despite
```julia
julia> @btime Arb(rand($BigFloat));
  271.410 ns (5 allocations: 336 bytes)
```

Joel-Dahne:
With #108 we now have
```julia
julia> @btime ArbMatrix(rand(Arb(prec=256), 100,100));
  2.647 ms (50007 allocations: 3.21 MiB)

julia> @btime ArbMatrix(rand(Arb, 100,100));
  2.657 ms (50006 allocations: 3.21 MiB)
```
So they are equal!

We also have
```julia
julia> @btime ArbMatrix(rand($BigFloat, 100, 100));
  1.827 ms (30006 allocations: 2.06 MiB)

julia> @btime ArbMatrix(Arb.(rand($BigFloat, 100, 100)));
  1.988 ms (50008 allocations: 3.28 MiB)

julia> @btime ArbMatrix(rand($Arb, 100, 100));
  2.895 ms (50006 allocations: 3.21 MiB)
```
It's natural that the first one is faster since it avoids one copy of each `Arb`. The two last ones have the same number of allocations but they are not equally fast, I'm not entirely sure why this is the case.

Notice that we still have the same problems as before with `Acb`.

kalmarek:
very well, thanks!

PR #108, opened 2020-12-21
Make Arf and Arb subtypes of Abstractfloat

Joel-Dahne:
I made `Arb` and `Acb` subtypes of `AbstractFloat`, this made some tests fail and I have tried to fix them as well. There are three things I have fixed:
- Adding a `Arb(::Rational)` for disambiguation
- Slightly change `Random.Sample`, it works slightly different for `AbstractFloat`. It should not have affected any functionality
- Rewrite the promotion system. This is the one change I have made which changes functionality, more comments about it below

I have factored out the promotions to `promote.jl` and changed them slightly. I have differentiated between promotion of internal types and external types. For the internal types I feel quite confident, we have a clear ordering of types `Mag < Arf < Arb < Acb` so it's clear how the promotions should work. `Ref`-types are promoted to non-ref types. For external types it's not a trivial what to do. Now I have implemented the following promotions
```julia
Base.promote_rule(::Type{<:ArfOrRef}, ::Type{<:Real}) = Arf
Base.promote_rule(::Type{<:ArbOrRef}, ::Type{<:Real}) = Arb
Base.promote_rule(::Type{<:AcbOrRef}, ::Type{<:Number}) = Acb
```
which I think makes sense in most cases, `Arf`, `Arb` and `Acb` are more general than any built in type in Julia at least. For `Mag` I'm not sure, the type is very limited and doesn't even convert `Float64´ properly, for now I have only implemented
```julia
Base.promote_rule(::Type{<:MagOrRef}, ::Type{<:Base.GMP.CdoubleMax}) = Mag
```
but maybe integers should be added as well? I have also added 
```julia
Base.promote_rule(::Type{<:ArbOrRef}, ::Type{<:Complex}) = Acb
```
so that `Arb` is promoted to `Acb` when used with other complex types. In hindsight changing the promotion should maybe have been done in a different pull request.

### DIFF
--- a/src/Arblib.jl
+++ b/src/Arblib.jl
@@ -49,6 +49,7 @@ include("setters.jl")
 include("constructors.jl")
 include("predicates.jl")
 include("show.jl")
+include("promotion.jl")
 include("arithmetic.jl")
 include("rand.jl")
 
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -1,24 +1,3 @@
-Base.promote_rule(::Type{<:MagLike}, ::Type{<:Union{Float64,MagLike}}) = Mag
-Base.promote_rule(::Type{<:ArfLike}, ::Type{<:Union{AbstractFloat,Integer,MagLike}}) = Arf
-Base.promote_rule(
-    ::Type{<:ArbLike},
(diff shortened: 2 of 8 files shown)


kalmarek (approved):
It all looks good to me;
Could you check what is the influence on #94 ?

Joel-Dahne:
It helps with that! See my comment in that issue

PR #111, opened 2020-12-23
Update Arb_jll and Flint_jll to version 2.19 and 2.7

Joel-Dahne:
This updates the versions used for Arb and Flint. The versioning number used in Julia has changed since the last release. [Here](https://github.com/JuliaPackaging/Yggdrasil/commit/1dc032884d96731f436f947c0753af4f3b25cdb8#diff-3f11581a12a8c59f140797f715d880bb90d8a70ab81db84768bf8ee5fcde26c4) we can read
```julia
# Arb_jll versions are decoupled from the upstream versions.
# Whenever we package a new official Arb release, we initially map its
# version X.Y.Z to X00.Y00.Z00 (i.e., multiply each component by 100).
# So for example version 2.19.0 would become 200.1900.000.
#
# Moreover, all our packages using Arb_jll use `~` in their compat ranges.
#
# Together, this allows us to increment the patch level of the JLL for minor tweaks.
# If a rebuild of the JLL is needed which keeps the upstream version identical
# but breaks ABI compatibility for any reason, we can increment the minor version
# e.g. go from 200.600.300 to 200.601.300.
# To package prerelease versions, we can also adjust the minor version; e.g. we may
# map a prerelease of 2.7.0 to 200.690.000.
#
# There is currently no plan to change the major version (except when Arb itself
# changes its major version. It simply seemed sensible to apply the same transformation
# to all components.
```
So the new version are `200.1900.0` and `200.700.0`. Now I have pinned those version exactly, but I'm not sure that is the best idea. Then we would have to make a new release every time a fix in `Arb_jll` or `Flint_jll` comes up, which seems unnecessary. We could allow the patch version to change I think, which would be
```
Arb_jll = "~200.1900"
FLINT_jll = "~200.700"
```
Then we are sure to use the same version of Arb but maybe a different packaging in Julia. What do you think?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,6 +11,6 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
-Arb_jll = "=2.18.1"
-FLINT_jll = "=2.6.3"
+Arb_jll = "~200.1900"
+FLINT_jll = "~200.700"
 julia = "1.3"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "0.2.2"
+version = "0.3.0"
 
 [deps]
 Arb_jll = "d9960996-1013-53c9-9ba4-74a4155039c3"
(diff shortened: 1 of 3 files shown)


kalmarek:
I'd be in favor of `~` versions;

Also there is a test in eigen that was leading to segfault on `2.18`, could you have a look at it?
https://github.com/kalmarek/Arblib.jl/blob/3e63f6637c387a8a806f8e8c85e89c76152c4aef/test/eigen.jl#L41

Joel-Dahne:
I switched to using `~` now. Indeed the segfault is fixed, I have reenabled those tests now.

kalmarek (approved):
maybe also version bump + release?

Joel-Dahne:
I bumped to version 0.3 now since #108 is possibly breaking. This will also close a number of issues so I can make the release after fixing that, so that it's correctly included in the release notes.

PR #112, opened 2020-12-25
Enh/improve mag support

Joel-Dahne:
This improves the support for `Mag`.
- Add construction from `Integer` by conversion to `UInt`
- Add `Mag(x::Integer, y::Integer)` returning `x*2^y`.
- Add support for more methods in `arithmetic.jl`, almost everything that `Mag` supports is added now
- Slightly rewrite some tests


src/arithmetic.jl
```diff
@@ -1,17 +1,34 @@
...
+Base.exp(x::MagOrRef) = exp!(zero(x), x)
+expinv(x::MagOrRef) = expinv!(zero(x), x)
+Base.expm1(x::MagOrRef) = expm1!(zero(x), x)
+Base.atan(x::MagOrRef) = atan!(zero(x), x)
+Base.cosh(x::MagOrRef) = cosh!(zero(x), x)
+Base.sinh(x::MagOrRef) = sinh!(zero(x), x)
+
+Base.min(x::MagOrRef, y::MagOrRef) = min!(zero(x), x, y)
+Base.max(x::MagOrRef, y::MagOrRef) = max!(zero(x), x, y)
+Base.minmax(x::MagOrRef, y::MagOrRef) = (min(x, y), max(x, y))
```
kalmarek, line 29: do we want to write those functions like this of through `@eval` in a loop (with 5 exceptions?)

kalmarek:
This looks good! Did you think about making `Mag<:Unsigned`, or is it asking too much?

Joel-Dahne:
`Unsigned` is for integers only, so that wouldn't make to much sense! :)

src/arithmetic.jl
Joel-Dahne: Maybe a loop will look nicer, strictly more lines but easier to change. Will fix that!

kalmarek:
:facepalm: of course, I'm dumb today :p

Joel-Dahne:
I happened to get an 8 hour train ride today with not much more to do, so I'm currently improving the whole `arithmetic.jl` :)

src/arithmetic.jl
Joel-Dahne: Actually it fit on three lines, so much nicer!

kalmarek (approved):

PR #114, opened 2020-12-26
Enh/improve arithmetic support

Joel-Dahne:
This pull request rewrites most of `artihmetic.jl`. It adds new methods, removes some which never existed, and adds tests for all of them. It also implements `minimum` and `maximum` and in relation to this I added a section to the README about common pitfalls when working with the Julia ecosystem, the issues with the default implementations of `minimum` and `maximum` are mentioned there.

The tests are failing for now but should start to work once #113 is merged and this is rebased on top of that.

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -101,6 +101,13 @@ end
 Base.:(^)(x::AcbOrRef, y::Union{AcbOrRef,ArbOrRef,_BitInteger}) =
     pow!(Acb(prec = _precision((x, y))), x, y)
 
+# We define the same special cases as Arb does, this avoids some
+# overhead
+Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{-2}) =
+    (y = inv(x); sqr!(y, y))
+#Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{-1}) - implemented in Base.intfuncs.jl
+Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{0}) = one(x)
+Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{1}) = copy(x)
 Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{2}) = sqr(x)
 
 Base.hypot(x::ArbOrRef, y::ArbOrRef) = hypot!(Arb(prec = _precision((x, y))), x, y)
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -180,7 +180,35 @@ Base.abs(z::AcbLike) = abs!(Arb(prec = _precision(z)), z)
 ### min and max
 for T in [MagOrRef, ArfOrRef, ArbOrRef]
     for op in [:min, :max]
+        # min/max
         @eval Base.$op(x::$T, y::$T) = $(Symbol(op, :!))(zero(x), x, y)
+        # minimum/maximum
+        # The default implementation of minimum/maximum in Julia is
+        # not correct for Arb, we implemented a correct version when
+        # no specific dimension is given. The default give correct
+        # results for Mag and Arf but wrong results for Arb in some
+        # cases, be careful!
+        @eval function Base.$(Symbol(op, :imum))(A::AbstractArray{<:$T})
+            isempty(A) &&
+                throw(ArgumentError("reducing over an empty collection is not allowed"))
+            res = copy(first(A))
+            for x in Iterators.drop(A, 1)
+                $(Symbol(op, :!))(res, res, x)
+            end
+            return res
+        end
     end
     @eval Base.minmax(x::$T, y::$T) = (min(x, y), max(x, y))
+
+    @eval function Base.extrema(A::AbstractArray{<:$T})
+        isempty(A) &&
+            throw(ArgumentError("reducing over an empty collection is not allowed"))
+        l = copy(first(A))
+        u = copy(first(A))
+        for x in Iterators.drop(A, 1)
+            min!(l, l, x)
+            max!(u, u, x)
+        end
+        return (l, u)
+    end
 end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -1,3 +1,7 @@
+const _BitInteger = (Int == Int64 ? Base.BitInteger64 : Base.BitInteger32)
+const _BitSigned = (Int == Int64 ? Base.BitSigned64 : Base.BitSigned32)
+const _BitUnsigned = (Int == Int64 ? Base.BitUnsigned64 : Base.BitUnsigned32)
+
 ### Mag
 for (jf, af) in [(:+, :add!), (:-, :sub!), (:*, :mul!), (:/, :div!)]
     @eval Base.$jf(x::MagOrRef, y::MagOrRef) = $af(zero(x), x, y)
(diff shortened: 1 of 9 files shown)


kalmarek:
This looks very good! thanks for taking care of this!

I was wondering about that `Union{Int, UInt}` argument. Technically everything that converts to `(U)Int` should be fine, so maybe we should use something along `Base.BitInteger64`/`Base.BitInteger32`?? (We can't handle `Base.BitInteger` since it includes e.g. `Int128`...). Maybe we could define our own
```julia
const _BitInteger = (Int == Int64 ? Base.BitInteger64 : Base.BitInteger32)
```
and use it in this context?

Joel-Dahne:
I got some free time over the holidays with all the travelling and such!

The `Int/UInt` idea is excellent! I looked in `Base` for something like that but didn't find it. Now I put the `_BitInteger` in the `arithmetic.jl` file, I was not sure where else to place it.

src/arithmetic.jl
```diff
@@ -18,63 +22,198 @@ for f in [:inv, :sqrt, :log, :log1p, :exp, :expm1, :atan, :cosh, :sinh]
...
+
+Base.:(-)(x::Union{ArbOrRef,AcbOrRef}) = neg!(zero(x), x)
+Base.abs(x::ArbOrRef) = abs!(zero(x), x)
+Base.:(/)(x::_BitUnsigned, y::ArbOrRef) = ui_div!(zero(y), x, y)
+
+# TODO: Should we convert Int to UInt for performance reasons?
+Base.:(^)(x::ArbOrRef, y::Union{ArbOrRef,_BitUnsigned}) =
+    pow!(Arb(prec = _precision((x, y))), x, y)
+Base.:(^)(x::AcbOrRef, y::Union{AcbOrRef,ArbOrRef,_BitInteger}) =
+    pow!(Acb(prec = _precision((x, y))), x, y)
```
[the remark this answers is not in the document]
kalmarek: since we're writing these methods directly I'd
```julia
Base.:(^)(x::ArbOrRef, y::ArbOrRef) = 
    pow!(Arb(prec = _precision((x, y))), x, y)
function Base.:(^)(x::ArbOrRef, y::_BitInteger)
    x, n = (y >= 0 ? (x, y) : (inv(x), abs(y)))
    return pow!(Arb(prec = precision(x)), convert(UInt, n))
end
Base.:(^)(x::AcbOrRef, y::Union{AcbOrRef,ArbOrRef,_BitInteger}) =
    pow!(Acb(prec = _precision((x, y))), x, y)

File: test/arithmetic.jl [hunk 2]
```diff
@@ -28,74 +28,204 @@
...
-        m[] = 1.0
-        @test Float64(m) ≥ 1.0
-        @test sprint(show, m) == sprint(show, m[])
+    @testset "Arb - specific" begin
+        @test Arb(1) + Arf(2) == Arf(2) + Arb(1) == Arb(3)
+        @test Arb(1) - Arf(2) == Arb(-1)
+        @test Arb(2) * Arf(3) == Arf(3) * Arb(2) == Arb(6)
+        @test Arb(6) / Arf(2) == UInt(6) / Arb(2) == UInt8(6) / Arb(2) == Arb(3)
+
+        @test Arb(2)^Arb(2) == Arb(2)^UInt(2) == Arb(2)^UInt8(2) == Arb(4)
```
kalmarek, line 155: also add
```julia
@test Arb(2)^2 == 2^2 == (Arb(2)^-2)^-1
```

test/arithmetic.jl
```diff
@@ -28,74 +28,204 @@
...
+        @test Acb(2) * Arb(3) == Arb(3) * Acb(2) == Acb(6)
+        @test Acb(6) / Arb(2) == Acb(3)
+
+        @test Acb(2)^Acb(2) ==
+              Acb(2)^Arb(2) ==
+              Acb(2)^2 ==
+              Acb(2)^Int8(2) ==
+              Acb(2)^UInt(2) ==
+              Acb(2)^UInt8(2) ==
+              Acb(4)
```
kalmarek, line 206:
```julia
(Acb(2)^-2)^-1 ==
```

kalmarek (changes requested):

test/arithmetic.jl
Joel-Dahne: Done! Well, except for the 2^2, where I just put 4 ^^

test/arithmetic.jl
Joel-Dahne: Done!

src/arithmetic.jl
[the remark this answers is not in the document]
Joel-Dahne: I a slightly different version of this that avoids one allocation by using `inv!`. I don't think we have to extend `Base.literal_pow` directly, the reason the first definition is not used in your example is that `-10` is not a `UInt` so it falls back to the default power by squaring. It works now
```julia
julia> Arb(2)^-10
0.00097656250000000000000000000000000000000000000000000000000000000000000000000000
```
We could possibly extend `Base.literal_pow` to use `Arblib.sqr` for `^2`. It seems to be about 10% faster actually when I try it.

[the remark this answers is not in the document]
Joel-Dahne: I extended `Base.literal_pow` for squaring now! I realized that `arb_pow_ui` actually starts by converting the `ui` to `fmpz` so this always require an extra allocation even in the cases which are later optimized, which are exponents `-2, -1, 0, 1, 2`. Adding a literal pow for all of these would avoid one allocation, at the moment `-1` and `2` are implemented.

src/arithmetic.jl
```diff
@@ -18,63 +22,203 @@ for f in [:inv, :sqrt, :log, :log1p, :exp, :expm1, :atan, :cosh, :sinh]
...
+Base.:(^)(x::ArbOrRef, y::ArbOrRef) = pow!(Arb(prec = _precision((x, y))), x, y)
+function Base.:(^)(x::ArbOrRef, y::_BitInteger)
+    z = zero(x)
+    x, n = (y >= 0 ? (x, y) : (inv!(z, x), -y))
+    return pow!(z, x, convert(UInt, n))
 end
+Base.:(^)(x::AcbOrRef, y::Union{AcbOrRef,ArbOrRef,_BitInteger}) =
+    pow!(Acb(prec = _precision((x, y))), x, y)
+
+Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{2}) = sqr(x)
```
kalmarek, line 104: let's go full in ;)
```suggestion
Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{2}) = sqr(x)
Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{1}) = copy(x)
Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{0}) = one(x)
# Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{-1}) = inv(x) : implemented in Base.intfuncs.jl
Base.literal_pow(::typeof(^), x::Union{ArbOrRef,AcbOrRef}, ::Val{-2}) = (res = sqr(x); inv!(res, res))
```

kalmarek (approved):

src/arithmetic.jl
Joel-Dahne: Done! It's a bit complicated to debug these things since `@which, @less, @enter` all makes it not use `literal_pow` anymore. But with some debug printing and performance comparisons I'm fairly certain they are actually used.

Joel-Dahne:
I realized I never did the 3.0.0 release. I can do it after merging this, or will that lead to any issues?

saschatimme:
No should be fine :)

PR #116, opened 2021-01-30
Add eps

Joel-Dahne:
This adds a `eps` method of `Arf`, `Arb` and `Acb`. I'll see if I can add one for `Mag` as well before merging this.

### DIFF
--- a/src/float.jl
+++ b/src/float.jl
@@ -1,6 +1,20 @@
-Base.eps(T::Type{<:Union{ArfOrRef,ArbOrRef,AcbOrRef}}) = eps(one(T))
-Base.eps(x::ArfOrRef) = Arf(set_ulp!(Mag(), x, prec = precision(x)), prec = precision(x))
-Base.eps(x::ArbOrRef) =
-    Arb(set_ulp!(Mag(), midref(x), prec = precision(x)), prec = precision(x))
-Base.eps(x::AcbOrRef) =
-    Acb(set_ulp!(Mag(), midref(realref(x)), prec = precision(x)), prec = precision(x))
+function eps!(res::ArfOrRef, x::ArfOrRef)
+    # FIXME: Should we return NaN here?
+    isspecial(x) && throw(ArgumentError("eps not defined for special values, got $x"))
+    return set!(res, set_ulp!(Mag(), x, prec = precision(x)))
+end
+
+function eps!(res::ArbOrRef, x::ArbOrRef)
+    # FIXME: Should we return NaN here?
+    mid_x = midref(x)
+    isspecial(mid_x) && throw(ArgumentError("eps not defined for special values, got $x"))
+    rad_res = radref(res)
+    set_ulp!(rad_res, mid_x, prec = precision(x))
+    return set!(res, rad_res)
+end
+
+function Base.eps(T::Type{<:Union{ArfOrRef,ArbOrRef}})
+    res = one(T)
+    return eps!(res, res)
+end
+Base.eps(x::Union{ArfOrRef,ArbOrRef}) = eps!(zero(x), x)
--- a/src/float.jl
+++ b/src/float.jl
@@ -1,0 +1,6 @@
+Base.eps(T::Type{<:Union{ArfOrRef,ArbOrRef,AcbOrRef}}) = eps(one(T))
+Base.eps(x::ArfOrRef) = Arf(set_ulp!(Mag(), x, prec = precision(x)), prec = precision(x))
+Base.eps(x::ArbOrRef) =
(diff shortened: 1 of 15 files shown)


Joel-Dahne:
I was confused because when I ran `JuliaFormatter` locally it changed some things which was already there. It seems that my local version (`0.12.3`) gives slightly different results than the version we are using `0.7.1`). Should we maybe update the version used for the package? Should I do that in this branch?

src/float.jl
```diff
@@ -0,0 +1,6 @@
+Base.eps(T::Type{<:Union{ArfOrRef,ArbOrRef,AcbOrRef}}) = eps(one(T))
+Base.eps(x::ArfOrRef) = Arf(set_ulp!(Mag(), x, prec = precision(x)), prec = precision(x))
+Base.eps(x::ArbOrRef) =
```
kalmarek, line 3: I think we could save one allocation here (at the cost of readability):
```julia
function eps(x::ArbOrRef)
    prec= precision(x)
    out = Arb(prec=prec)
    rad = radref(out)
    set_ulp!(rad, midref(x), prec=prec)
    set!(midref(out), rad)
    Arblib.zero!(rad)
    return out
end
```
with benchmarks for `x=Arb(rand())`:
```julia
BenchmarkTools.Trial: # old eps
  memory estimate:  128 bytes
  allocs estimate:  3
  --------------
  minimum time:     72.061 ns (0.00% GC)
  median time:      75.540 ns (0.00% GC)
  mean time:        301.690 ns (44.43% GC)
  maximum time:     163.718 μs (80.91% GC)
  --------------
  samples:          10000
  evals/sample:     968
```
```julia
BenchmarkTools.Trial: # new eps
  memory estimate:  96 bytes
  allocs estimate:  2
  --------------
  minimum time:     47.383 ns (0.00% GC)
  median time:      49.274 ns (0.00% GC)
  mean time:        169.549 ns (43.90% GC)
  maximum time:     134.622 μs (84.70% GC)
  --------------
  samples:          10000
  evals/sample:     984
```

as for the (mathematical) definition, I'm fine with what is here;
I tried to solicit some advice on this, but got cryptic/no answers, see
https://discourse.julialang.org/t/doublefloats-jl-double128-quad64/51049/55?u=abulak

kalmarek (changes requested):

kalmarek:
let's update the JuliaFormatter we use in actions, shall we? @saschatimme

saschatimme:
> let's update the JuliaFormatter we use in actions, shall we? @saschatimme

sure

Joel-Dahne:
I added an inplace version of `eps` and put the complicated code there. If people really want to compute their epsilons quick we should not stop them! I also fixed some issues with aliasing I encountered while implementing this. I removed `eps` for `Acb` since I noticed that Julia doesn't implement it for `Complex` and it's unclear how we would handle the real and imaginary part.

I realized that `set_ulp!` panics for special values. Now I made it throw an error instead. It might be a better alternative to make it return `NaN` instead, this is what Julia does for `Inf`. One noticeable difference to `Float64` is that `eps` of zero is not well defined for `Arf` due to the arbitrary precision exponent.  

I still want to add a version for `Mag` but I'm not entirely sure how. My current idea is `Base.eps(x::MagOrRef) = mul_2exp!(Mag(), x, -30)`, where 30 comes from `Mag` having 30 bits of precision. I believe this is correct for `Mag(1)` but otherwise I'm not sure.

kalmarek (approved):

kalmarek:
unless You really need it for Mag, I'd merge this and leave Mag one as an issue; eps is the first thing most numerical Base stuff will fail with Arblib

Joel-Dahne:
Sounds reasonable! What do you think about how to handle special values for `Arf`? Throw an error or return `NaN`?

kalmarek:
I think throwing an error is a better option; at least you will know where these Nans would have came from...

Joel-Dahne:
After sleeping on it I think I changed my mind. I think it's more consistent with Julia to return `NaN`. It would be confusing if `eps(Inf) = NaN`  but `eps(Arb(Inf))` throws an error, there is no reason why they should differ. I think having `eps(zero(Arb)) = NaN` would be the most consistent as well. I have been bit several times by running a long computation on several different inputs just to have it throw an error towards the end instead of returning `NaN` for that input, making me lose all the other computations as well, so in general I prefer return `NaN` instead of throwing when it makes sense.

kalmarek:
@Joel-Dahne sure, feel free to merge then!

Joel-Dahne:
I realized we didn't have a functioning `isnan` for `Arb` and `Acb`, so I fixed that as well! Then I'll merge this soon!

PR #117, opened 2021-02-18
Enh/special functions

Joel-Dahne:
I have added an interface to [SpecialFunctions.jl](https://github.com/JuliaMath/SpecialFunctions.jl/). All of the code is in `special-functions.jl` where I have written down all the special functions implemented by `SpecialFunction`. For each one I have either given a corresponding Arb implementation or written a small comment about why that one is not handled. Some thinks I though about when doing this:

- How to handle scaled version not implemented by Arb? Should we just make a naive implementation or skip it? If we do implement them the user might assume that they are optimized for the scaled version, on the other hand the user will always get a bound on the error so knows if more precision is needed. If you write generic code with the scaled version it would be nice if it just worked. Maybe we should talk to the developers of `SpecialFunctions` about implementing fallback versions for generic scaled functions?
- Many of the functions take more than one argument, which one should we take the precision from? In most other cases we have taken the maximum of all the inputs so I guess that makes sense here as well? The current implementation doesn't take this into account.
- Should we handle promotion in any way? It would probably be better if this was implemented in `SpecialFunctions` directly.
- Some (which?) Arb functions are more general than those in `SpecialFunctions`. How should we handle that?
- I have not checked that the branch cuts are the same for `SpecialFunctions` and Arb, I believe in most cases they are. 
- Should we add documentation for the methods which are not identical to those in SpecialFunctions (which are these?)? Should we just refer to Arbs documentation?

Do you have any thoughts about these?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "0.3.0"
+version = "0.3.1"
 
 [deps]
 Arb_jll = "d9960996-1013-53c9-9ba4-74a4155039c3"
--- a/Project.toml
+++ b/Project.toml
@@ -8,9 +8,11 @@ Arb_jll = "d9960996-1013-53c9-9ba4-74a4155039c3"
 FLINT_jll = "e134572f-a0d5-539d-bddf-3cad8db41a82"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 Arb_jll = "~200.1900"
 FLINT_jll = "~200.700"
 julia = "1.3"
+SpecialFunctions = "1.0"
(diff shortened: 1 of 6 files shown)


kalmarek:
That is some really meticulous job! My thoughts on the raised points:
* when Arb doesn't implement scaled functions I'd say we should provide fallback; generic fallbacks will not work for Arblib for performance reasons
* I'm fine with the maximum of precisions
* What kind of promotion do you mean? In general I'd stay away from promotion rules as these will only complicate handling precision
* If Arblib functions are more general, I'd leave it to the user to look up the Arblib docs. It's hard to imagine ( at least to me:) ) somebody writing code with low-level Arblib.jl without consulting the original docs. The initial purpose of this package was to provide a user friendly interface to Arb library, so I'd stay away from changing how Arb behaves. Of course we can always discuss the purposes and aims ;), but I'm slightly afraid that producing a user-friendly package for general numerical computations in julia based on Arb needs more time/devotion than I can provide ;)

Would be great to hear @saschatimme take on this.

Joel-Dahne:
Thanks!

- That sounds reasonable I think. I will look at implementing the scaled versions then. 
- Then I'll make sure that all methods use the maximum of the precision.
- For example for `besselj(ν, z)` it's quite common that `ν` is an integer and people might want to call it like `besselj(1::Int, z::Arb)`. But I think we should let this be handled by `SpecialFuntions` and not by us.
- I agree, we can write a note in the documentation referring to the Arb documentation for more details.

I happened to find some time to create this pull request but otherwise I don't have to much time at the moment, so might be a while before I get everything done!

Joel-Dahne:
I did some updates now. More specifically I rewrote the methods to minimize the number of allocations (only one of them allocates things than it returns now) and to use as precision the maximum of the inputs. I also added some methods which I realized I had missed. Finally I added a small section about special functions to the README.

I have not added any scaled versions or versions otherwise not implemented by Arb. It might be something we want to do in the future but I feel like it's not required for this first version.

I'm ready to merge if you are happy with how it looks now!

Joel-Dahne:
I also bumped it to version 0.3.1 now.

kalmarek (approved):
Looks great ;)

PR #121, opened 2021-04-21
Add methods related to intervals

Joel-Dahne:
I have implemented a bunch of interval related methods, these are
- radius
- midpoint
- lbound
- ubound
- abs_lbound
- abs_ubound
- getinterval
- getball
- union
- intersection
Most of them are self explanatory but I have also added documentation for them.

I have three thoughts/questions related to this
1. Which of these to export? Now I have exported all of them but maybe we should not do that?
2. What type to return? Now I have made it so that by default the methods return the most basic type, for example `Mag` for `radius` and `Arf` for `lbound`, but they also support conversion to other types with for example `lbound(Arb, x)` (similar to how `ceil` works for example). In the beginning I had them always return `Arb` by default but I changed my mind so now they return the most basic type. I think it's better to be explicit when you want it converted to `Arb`.
3. What to name intersection? Now `union` uses `Base.union` however for `intersection` there is no such base-method, instead the base method is called `intersect`. Should we keep calling it `intersection` and export that or change to using `Base.intersect`?

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -88,7 +88,7 @@ end
 abs_ubound(x::Union{ArbOrRef,AcbOrRef}) = abs_ubound(Arf, x)
 
 """
-    getinterval([T, ] x::Union{ArbOrRef,AcbOrRef})
+    getinterval([T, ] x::ArbOrRef)
 
 Returns a tuple `(l, u)` representing an interval `[l, u]` enclosing
 the ball `x`, both of them are of type `Arf`. If `T` is given convert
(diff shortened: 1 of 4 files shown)


Joel-Dahne:
Okay `contains` doesn't exist in 1.3 and 1.4 apparently. I can fix that later. Maybe we should start testing on 1.6 as well?

src/interval.jl
```diff
@@ -0,0 +1,168 @@
...
+"""
+function intersection(x::ArbOrRef, y::ArbOrRef)
+    overlaps(x, y) ||
+        throw(ArgumentError("intersection of non-intersecting balls not allowed"))
+    res = Arb(prec = _precision((x, y)))
+    intersection!(res, x, y)
+    return res
+end
+# TODO: Could be optimized, both for performance and enclosure
+intersection(x::ArbOrRef, y::ArbOrRef, z::ArbOrRef, xs...) =
```
kalmarek, line 167: IntervalArithmetic uses `Base.intersect`. maybe we could define it referring to `intersection`?
https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/e0333aee5eecc90525bd1d0b9536e984dd248ab7/src/intervals/set_operations.jl#L74

kalmarek (changes requested):

kalmarek:
I'm fine with 1 and 2;
with 3 I'd go for `Base.intersect` and maybe provide `intersection` internally (not exported).

Joel-Dahne:
I renamed `intersection` to `intersect` and updated the documentation (not exactly like you worded it but similar). I also fixed the issue with `contains` not being exported before Julia 1.5 and removed some TODOs I had forgotten in the test file. I have not added an internal `intersection` method, it would make it easier to switch between `intersection` and `intersection!` however.

kalmarek (approved):
lgtm!

PR #123, opened 2021-05-15
Scalar multiplication and division with matrices

### LINKED ISSUE #122: Arb * ArbMatrix becomes Matrix{Arb} instead of ArbMatrix
Hi, I came across this issue a few times now, so I thought I would mention it here.
```
julia> Arb(2)*ArbMatrix([1 2; 3 4])
2×2 Matrix{Arb}:
 2.0000000000000000000000000000000000000000000000000000000000000000000000000000  …  4.0000000000000000000000000000000000000000000000000000000000000000000000000000
 6.0000000000000000000000000000000000000000000000000000000000000000000000000000     8.0000000000000000000000000000000000000000000000000000000000000000000000000000
```
So after multiplying a scalar with a matrix things start to break down because you don't have an `ArbMatrix` anymore
My workaround is using `Arblib.addmul!(B,A,c)` or `Arblib.mul!(B,A,c)` depending on the case, but I think this should be added here too because it is a common operation for matrices.  I'm not sure about the types (e.g., how to handle cases like Acb * ArbMatrix except for defining separate functions for them), but without typing something like this would probably suffice:
```
function Base.:(*)(c, A)
    B = similar(A)
    Arblib.mul!(B,A,c)
    B
end
```

nanleij:
Added `c*A`, `A*c`, `c\A` and `A/c` and some tests. This would close #122 
I am not sure about `Acb * ArbMatrix` (now I did, but I use explicit conversion to an `AcbMatrix`), maybe explicit conversion from the user is better there?

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -170,7 +170,7 @@ end
 
 # scalar multiplication
 function Base.:(*)(c::ArbLike, A::T) where {T <: Matrices}
-    C = T(size(A,1),size(A,2),prec = precision(A))
+    C = similar(A)
     Arblib.mul!(C,A,c)
 end
 function Base.:(*)(c::AcbLike, A::AcbMatrixLike)
@@ -179,14 +179,14 @@ function Base.:(*)(c::AcbLike, A::AcbMatrixLike)
 end
 function Base.:(*)(c::AcbLike, A::ArbMatrixLike)
     C = AcbMatrix(A)
-    Arblib.mul!(C,AcbMatrix(A),c)
+    Arblib.mul!(C,C,c)
 end
 Base.:(*)(A::T, c::AcbLike) where {T <: Matrices} = c*A
 Base.:(*)(A::T, c::ArbLike) where {T <: Matrices} = c*A
 
 # scalar division
 function Base.:(\)(c::ArbLike, A::T) where {T <: Matrices}
-    C = T(size(A,1),size(A,2),prec = precision(A))
+    C = similar(A)
     Arblib.div!(C,A,c)
 end
 function Base.:(\)(c::AcbLike, A::AcbMatrixLike)
@@ -195,7 +195,7 @@ function Base.:(\)(c::AcbLike, A::AcbMatrixLike)
 end
 function Base.:(\)(c::AcbLike, A::ArbMatrixLike)
     C = AcbMatrix(A)
-    Arblib.div!(C,AcbMatrix(A),c)
+    Arblib.div!(C,C,c)
 end
 Base.:(/)(A::T, c::AcbLike) where {T <: Matrices} = c\A
 Base.:(/)(A::T, c::ArbLike) where {T <: Matrices} = c\A
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -169,36 +169,36 @@ function Base.:(*)(A::T, B::T) where {T<:Matrices}
 end
 
 # scalar multiplication
-function Base.:(*)(c::ArbLike, A::T) where {T <: Matrices}
+function Base.:(*)(c::ArbLike, A::T) where {T<:Matrices}
     C = similar(A)
-    Arblib.mul!(C,A,c)
+    Arblib.mul!(C, A, c)
 end
 function Base.:(*)(c::AcbLike, A::AcbMatrixLike)
     C = similar(A)
-    Arblib.mul!(C,A,c)
+    Arblib.mul!(C, A, c)
 end
 function Base.:(*)(c::AcbLike, A::ArbMatrixLike)
     C = AcbMatrix(A)
-    Arblib.mul!(C,C,c)
+    Arblib.mul!(C, C, c)
 end
-Base.:(*)(A::T, c::AcbLike) where {T <: Matrices} = c*A
-Base.:(*)(A::T, c::ArbLike) where {T <: Matrices} = c*A
+Base.:(*)(A::T, c::AcbLike) where {T<:Matrices} = c * A
+Base.:(*)(A::T, c::ArbLike) where {T<:Matrices} = c * A
 
 # scalar division
-function Base.:(\)(c::ArbLike, A::T) where {T <: Matrices}
+function Base.:(\)(c::ArbLike, A::T) where {T<:Matrices}
     C = similar(A)
-    Arblib.div!(C,A,c)
+    Arblib.div!(C, A, c)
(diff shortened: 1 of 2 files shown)


src/matrix.jl
```diff
@@ -168,6 +168,38 @@ function Base.:(*)(A::T, B::T) where {T<:Matrices}
...
+end
+function Base.:(*)(c::AcbLike, A::AcbMatLike)
+    C = similar(A)
+    Arblib.mul!(C,A,c)
+end
+function Base.:(*)(c::AcbLike, A::ArbMatrixLike)
+    C = AcbMatrix(A)
+    Arblib.mul!(C,AcbMatrix(A),c)
+end
+Base.:(*)(A::T, c::AcbLike) = c*A
```
kalmarek, line 184: `where T<:??`

kalmarek (changes requested):

src/matrix.jl
nanleij: Good point, changed that in the last commit

src/matrix.jl
```diff
@@ -168,6 +168,38 @@ function Base.:(*)(A::T, B::T) where {T<:Matrices}
     LinearAlgebra.mul!(C, A, B)
 end
 
+# scalar multiplication
+function Base.:(*)(c::ArbLike, A::T) where {T <: Matrices}
+    C = T(size(A,1),size(A,2),prec = precision(A))
```
Joel-Dahne, line 173: Here you could use `similar(A)`.

src/matrix.jl
```diff
@@ -168,6 +168,38 @@ function Base.:(*)(A::T, B::T) where {T<:Matrices}
...
+    Arblib.mul!(C,A,c)
+end
+function Base.:(*)(c::AcbLike, A::AcbMatrixLike)
+    C = similar(A)
+    Arblib.mul!(C,A,c)
+end
+function Base.:(*)(c::AcbLike, A::ArbMatrixLike)
+    C = AcbMatrix(A)
+    Arblib.mul!(C,AcbMatrix(A),c)
```
Joel-Dahne, line 182: This could be`Arblib.mul!(C, C, c)`, saving one allocation of the matrix.

src/matrix.jl
```diff
@@ -168,6 +168,38 @@ function Base.:(*)(A::T, B::T) where {T<:Matrices}
...
+end
+Base.:(*)(A::T, c::AcbLike) where {T <: Matrices} = c*A
+Base.:(*)(A::T, c::ArbLike) where {T <: Matrices} = c*A
+
+# scalar division
+function Base.:(\)(c::ArbLike, A::T) where {T <: Matrices}
+    C = T(size(A,1),size(A,2),prec = precision(A))
```
Joel-Dahne, line 189: Here you can also use `similar`.

src/matrix.jl
```diff
@@ -168,6 +168,38 @@ function Base.:(*)(A::T, B::T) where {T<:Matrices}
...
+    Arblib.div!(C,A,c)
+end
+function Base.:(\)(c::AcbLike, A::AcbMatrixLike)
+    C = similar(A)
+    Arblib.div!(C,A,c)
+end
+function Base.:(\)(c::AcbLike, A::ArbMatrixLike)
+    C = AcbMatrix(A)
+    Arblib.div!(C,AcbMatrix(A),c)
```
Joel-Dahne, line 198: Same as the comment above, you can use `Arblib.div!(C, C, c)`.

test/matrix.jl
```diff
@@ -44,6 +44,17 @@
...
+    
+    @testset "scalar arithmetic" begin
+        AInt = [2 4; 6 8]
+        A = TMat(AInt; prec = 96)
+        cInt = 2
+        c = T(cInt; prec= 96)
+        @test c*A isa TMat
+        @test c*A == cInt*AInt
+        @test c\A isa TMat
+        @test c\A == cInt\AInt
```
Joel-Dahne, line 56: Add tests for the reversed version as well!

test/matrix.jl
```diff
@@ -44,6 +44,17 @@
...
+    end
```
Joel-Dahne, line 57: To also try the promotion to `AcbMatrix` you could add
```julia
if TRef <: Real
    @test Acb(c) * A isa AcbMatrix
    @test Acb(c) * A == c * A
end
```
and also add `c \ A`, `A * c` and `A / c`.

Joel-Dahne (changes requested):
This looks great! Thank you! Allow me to do some nitpicking. Also, in the end you should run [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) on it!

nanleij:
Thanks for the feedback, I added your suggestions!

test/matrix.jl
```diff
@@ -44,6 +44,33 @@
...
+        if TRef <: Real
+            @test Acb(c) * A is AcbMatrix
+            @test Acb(c) * A == c * A
+            @test A * Acb(c) is AcbMatrix
+            @test A * Acb(c) == A * c
+
+            @test Acb(c) \ A is AcbMatrix
+            @test Acb(c) \ A == c \ A
+            @test A / Acb(c) is AcbMatrix
+            @test A / Acb(c) == A / c
```
Joel-Dahne, line 71: Should be `isa` instead if `is` here.

Joel-Dahne:
This looks great I think! Apart from the comment above I think it should be good to go! Do you have any more comments @kalmarek ?

Joel-Dahne:
The formatter seems to be unhappy. But I think that's at least partially because they have released a new version which is not backwards compatible. Could you try to format it again using version `0.12.3` of `JuliaFormatter`? You should be able to install it with `add JuliaFormatter @0.12.3`. Then we can update the formatter at a later stage.

nanleij:
Sure I'll do that right now

nanleij:
It should be okay now

Joel-Dahne (approved):

Joel-Dahne:
Looks good to me now! Anything more to add @kalmarek ?

kalmarek (approved):

kalmarek:
sorry for late response, I'm currently on holiday ;)
@Joel-Dahne merge and tag at will ;)

Joel-Dahne:
No worries! Enjoy! I'll merge this now and make another pull request for  updating the JuliaFormatter version and making a release.

PR #124, opened 2021-05-22
Update formatter and release

Joel-Dahne:
I updated the version used for the formatter as well as bumped the version to 0.3.2.


Joel-Dahne:
Thought it was a good idea to include Julia v1.6 in the CI as well!

saschatimme (approved):

kalmarek:
@Joel-Dahne should we tag on current master?

Joel-Dahne:
Yes that we could do!

PR #125, opened 2021-06-08
Improve support for polynomials and series

Joel-Dahne:
I have been working on this change for a long time and finally managed to get closer to finishing it. It's basically a completely rewrite of polynomials and series and also adds most of the functionality for them. There are still some things left to do, mainly adding an interface to `SpecialFunctions` for series.

I push it now in case you have any comments and to see if the tests pass on all different platforms. It's a very large diff but there is no hurry to get it done so don't feel like you have to go through it anytime soon.


kalmarek (approved):
This is wonderful!
I didn't have much time to go in detail through this, but lgtm ;)

Maybe the only thing that caught my attention was comments documenting missing features/issues in a few places. It'd be better to convert those to gh issues maybe?

Joel-Dahne:
I'll try to get the interface to SpecialFunctions done somewhat soon (we'll see...) and after that I will go through all the TODOs I have left so far. Indeed it might be a good idea to convert some of the to GH issues.

kalmarek:
did you try wrapping root finding? I made half-hearted attempt here:
https://gist.github.com/kalmarek/cf4b2b9be4d2300e6060018a60fed049

but I run into this: https://github.com/fredrik-johansson/arb/issues/367

Also some functions aren't exposed by Arb library (static inline) but I think we could do without them. I couldn't go around the segfault though...

kalmarek:
anyway, merge at will ;)

Joel-Dahne:
Okay, now I think I'm done with the pull request! I added support for methods from `SpecialFunctions` and some missing constructors mainly. I'll let it rest for a day or two in case you have some comments or I realize I missed something, then I'll merge it!

PR #130, opened 2021-09-09
Add ref method for polynomials and series

Joel-Dahne:
This finally allows for implementing non-allocating methods for polynomials and series.


kalmarek (approved):
lgtm!

Issue #131, opened 2021-09-16
How can we enable multi-threaded Flint via Arblib.jl?


lucifer1004:
Arb has a `flint_set_num_threads()` function to enable multi-threaded Flint, and Nemo.jl exposes this function, however, I do not see it in Arblib.jl.

kalmarek:
It seems that this is flints function, not arblibs;

There is an env flag that needs to be set before loading `Flint_jll`, see
https://github.com/JuliaBinaryWrappers/FLINT_jll.jl/blob/main/src/wrappers/x86_64-linux-gnu.jl
then we need to store the existence of this flag at `__init__` time and the thing will be a simple `ccall` to `libflint`.

@lucifer1004 mind making a PR?

Joel-Dahne:
Closed by #132

PR #132, opened 2021-09-16
feat: add support for multi-threading flint

lucifer1004:
Solves #131 

The code was copied from `Nemo.jl` with minor modifications (replacing `Nemo` with `Arblib`). The usage has been added to `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,3 +262,9 @@ and does are not extended, these are mostly scaled version of methods.
 Arb does however implement many special functions that are not in
 SpecialFunction and at the moment there is no user friendly interface
 for most of them.
+
+## Support for multi-threading
+Enabling a threaded version of flint can be done by setting the
+environment variable `NEMO_THREADED=1`. Note that this should be
+set before `Arblib.jl` is loaded. To set the actual number of threads,
+use `Arblib.flint_set_num_threads($numberofthreads)`.
--- a/src/Arblib.jl
+++ b/src/Arblib.jl
@@ -31,7 +31,8 @@ export Mag,
     ArbSeries,
     AcbPoly,
     AcbSeries,
-    ref
+    ref,
+    flint_set_num_threads
 
 macro libarb(function_name)
     return (:($function_name), libarb)
@@ -39,6 +40,20 @@ end
 
 macro libflint(function_name)
     return (:($function_name), Arb_jll.libflint)
+end
+
+const __isthreaded = Ref(false)
+
+function __init__()
+    __isthreaded[] = Base.get(ENV, "NEMO_THREADED", "") == "1"
+end
+
+function flint_set_num_threads(a::Integer)
+    if !__isthreaded[]
+        error("To use threaded flint, julia has to be started with NEMO_THREADED=1")
+    else
+        ccall(@libflint(flint_set_num_threads), Nothing, (Cint,), a)
+    end
 end
 
 include("arb_types.jl")
--- a/src/Arblib.jl
+++ b/src/Arblib.jl
@@ -31,8 +31,7 @@ export Mag,
     ArbSeries,
     AcbPoly,
     AcbSeries,
-    ref,
-    flint_set_num_threads
+    ref
 
 macro libarb(function_name)
     return (:($function_name), libarb)
--- a/Project.toml
+++ b/Project.toml
@@ -14,5 +14,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 Arb_jll = "~200.1900"
 FLINT_jll = "~200.700"
-julia = "1.3"
 SpecialFunctions = "1.0"
+julia = "1.3"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,3 +1,5 @@
+ENV["NEMO_THREADED"] = 1
+
 using Arblib, Test, LinearAlgebra, SpecialFunctions
 
 @testset "Arblib" begin
@@ -23,4 +25,5 @@ using Arblib, Test, LinearAlgebra, SpecialFunctions
     include("poly.jl")
     include("series.jl")
     include("special-functions.jl")
+    include("threading.jl")
 end
--- a/test/threading.jl
+++ b/test/threading.jl
@@ -1,0 +1,38 @@
+@testset "Threading" begin
+    @testset "Check number of threads" begin
+        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 1
+
+        Arblib.flint_set_num_threads(2)
+        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 2
+
+        Arblib.flint_set_num_threads(4)
+        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 4
+
+        Arblib.flint_set_num_threads(1)
+        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 1
+    end
+
+    @testset "Disabled threading" begin
+        # This manually unsets the threading flag and checks that we
+        # get an error when trying to set the number of threads
+        Arblib.__isthreaded[] = false
+
+        @test_throws ErrorException Arblib.flint_set_num_threads(2)
+
+        Arblib.__isthreaded[] = true
+    end
+
+    @testset "Matrix multiplication" begin
+        A = ArbMatrix(reshape(1:10000, 100, 100))
+
+        Arblib.flint_set_num_threads(1)
+        B1 = Arblib.mul_threaded!(similar(A), A, A)
+
+        Arblib.flint_set_num_threads(2)
+        B2 = Arblib.mul_threaded!(similar(A), A, A)
+
+        @test B1 == B2 == A * A
+
+        Arblib.flint_set_num_threads(2)
+    end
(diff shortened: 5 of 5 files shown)


lucifer1004:
It seems that we have to use `NEMO_THREADED` since it was hard-coded [here](https://github.com/JuliaBinaryWrappers/FLINT_jll.jl/blob/9a4e5a48246b4a7c4048ff1d2d574a30a972d705/src/wrappers/x86_64-linux-gnu.jl#L17).

src/Arblib.jl
```diff
@@ -41,6 +42,25 @@ macro libflint(function_name)
...
+function __init__()
+    __isthreaded[] = Base.get(ENV, "NEMO_THREADED", "") == "1"
+
+    if __isthreaded[]
+        ccall((:__gmp_set_memory_functions, :libgmp), Nothing,
+            (Int, Int, Int), 0, 0, 0)
+    end
+end
+
+function flint_set_num_threads(a::Int)
```
kalmarek, line 56:
```suggestion
function flint_set_num_threads(a::Integer)
```
anything that `cconverts` to `Clong` should be fine

src/Arblib.jl
```diff
@@ -41,6 +42,25 @@ macro libflint(function_name)
...
+    if !__isthreaded[]
+        error("To use threaded flint, julia has to be started with NEMO_THREADED=1")
+    else
+        ccall(@libflint(flint_set_num_threads), Nothing, (Int,), a)
```
kalmarek, line 60: is that a `Clong`, or `Cint` in flint?

src/Arblib.jl
```diff
@@ -41,6 +42,25 @@ macro libflint(function_name)
     return (:($function_name), Arb_jll.libflint)
 end
 
+const __isthreaded = Ref(false)
+
+function __init__()
+    __isthreaded[] = Base.get(ENV, "NEMO_THREADED", "") == "1"
+
+    if __isthreaded[]
+        ccall((:__gmp_set_memory_functions, :libgmp), Nothing,
```
kalmarek, line 51: why do we need to set those memory functions? also is that really `0::Int` or rather `C_NULL::Ptr{Cvoid}`?

kalmarek (changes requested):

src/Arblib.jl
lucifer1004: I just copied the codes from `Nemo.jl` codebase, since I am not familiar with `gmp` or `flint`.

src/Arblib.jl
kalmarek: It's actually `Cint`: https://flintlib.org/doc/flint.html?highlight=flint_set_num_threads#c.flint_set_num_threads
```suggestion
        ccall(@libflint(flint_set_num_threads), Nothing, (Cint,), a)
```

src/Arblib.jl
kalmarek: @Joel-Dahne could you chip in?

Joel-Dahne:
Thanks for taking a look at this, it would be great to get working! Let me take a look at it tomorrow!

src/Arblib.jl
Joel-Dahne: It works for me even if I don't set these memory functions, but I don't know what they do so it could depend on your system or on what you are doing for example. It actually works even without setting the `NEMO_THREADED` to `1` for me, so I don't even know what that part does...

It could maybe be that Nemo has these methods for historical reasons, they use Flint before `flint_jll` was a thing, and that they are not needed anymore.

How about we try to just not set these memory functions for now and hope that things work? And then we add it back if we notice any issues? I would like to avoid touching gmp things which I don't understand if we can avoid it.

lucifer1004: Are you suggesting we just wrap the `flint_set_num_threads` function and do nothing else?

README.md
```diff
@@ -29,18 +29,18 @@ base Julia. They contain a pointer to an object of the corresponding
...
+| `acb_t`      | `Acb` or `AcbRef`             |
+| `arb_t*`     | `ArbVector` or `ArbRefVector` |
+| `acb_t* `    | `AcbVector` or `AcbRefVector` |
 | `arb_mat_t`  | `ArbMatrix` or `ArbRefMatrix` |
-| `acb_mat_t`  | `AcbMatrix` or `AcbRefMatrix`  |
-| `arb_poly_t`  | `ArbPoly` or `ArbSeries` |
-| `acb_poly_t`  | `AcbPoly` or `AcbSeries` |
+| `acb_mat_t`  | `AcbMatrix` or `AcbRefMatrix` |
+| `arb_poly_t` | `ArbPoly` or `ArbSeries`      |
+| `acb_poly_t` | `AcbPoly` or `AcbSeries`      |
```
Joel-Dahne, line 43: Would it be possible to revert this change? Since it's not part of the threading.

src/Arblib.jl
```diff
@@ -5,10 +5,10 @@ import LinearAlgebra
 import SpecialFunctions
 
 if VERSION >= v"1.5.0-DEV.639"
-    import Base: contains
+    import Base:contains
 end
 if VERSION >= v"1.6"
-    import Base: sincospi
+    import Base:sincospi
```
Joel-Dahne, line 11: Would it be possible to reverse this as well? Again not part of the treading.

Joel-Dahne (commented):
Other than the comments above this looks good and seems to work well on my machine as well! I think it would be a good idea to add some tests for it as well. I would propose adding this
```julia
@testset "Threading" begin
    @testset "Check number of threads" begin
        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 1

        Arblib.flint_set_num_threads(2)
        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 2

        Arblib.flint_set_num_threads(4)
        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 4

        Arblib.flint_set_num_threads(1)
        @test ccall(Arblib.@libflint(flint_get_num_threads), Cint, ()) == 1
    end

    @testset "Disabled threading" begin
        # This manually unsets the threading flag and checks that we
        # get an error when trying to set the number of threads
        Arblib.__isthreaded[] = false

        @test_throws ErrorException Arblib.flint_set_num_threads(2)

        Arblib.__isthreaded[] = true
    end

    @testset "Matrix multiplication" begin
        A = ArbMatrix(reshape(1:10000, 100, 100))

        Arblib.flint_set_num_threads(1)
        B1 = Arblib.mul_threaded!(similar(A), A, A)

        Arblib.flint_set_num_threads(2)
        B2 = Arblib.mul_threaded!(similar(A), A, A)

        @test B1 == B2 == A * A

        Arblib.flint_set_num_threads(2)
    end
end
```
in a file `test/threading.jl` and then updating `test/runtest.jl` to be
```julia
ENV["NEMO_THREADED"] = 1

using Arblib, Test, LinearAlgebra, SpecialFunctions

@testset "Arblib" begin
    include("arb_types-test.jl")
    include("types-test.jl")
    include("arbcall-test.jl")
    include("precision-test.jl")
    include("setters-test.jl")
    include("constructors-test.jl")
    include("predicates-test.jl")
    include("show-test.jl")
    include("promotion.jl")
    include("examples.jl")
    include("arithmetic.jl")
    include("rand.jl")
    include("float.jl")
    include("interval.jl")
    include("vector.jl")
    include("matrix.jl")
    include("eigen.jl")
    include("calc_integrate.jl")
    include("ref-test.jl")
    include("poly.jl")
    include("series.jl")
    include("special-functions.jl")
    include("threading.jl")
end
```

src/Arblib.jl
Joel-Dahne: Exactly! At least it works on my Linux machine. I don't know what you run on but it would be good if you could test that it works as well. Also if we add the tests I propose below they will also be run on both Windows and Mac with the continuous integration.

Joel-Dahne:
Oh, and it seems like the formatter was complaining as well, you can format the code with
```julia
using JuliaFormatter
format(".")
```

lucifer1004:
@Joel-Dahne @kalmarek Please review the latest commit.

Joel-Dahne:
This looks good to me! Great job! Lets see if @kalmarek has anything to add as well. Normally I would propose to push the version as well, to make a new release. However I plan to push some other changes later today or tomorrow, so we can maybe wait with the new release until then.

src/Arblib.jl
```diff
@@ -31,7 +31,8 @@ export Mag,
     ArbSeries,
     AcbPoly,
     AcbSeries,
-    ref
+    ref,
+    flint_set_num_threads
```
kalmarek, line 35: we shouldn't export this function

kalmarek (commented):
other than exporing `flint_set_num_threads` this looks good!

kalmarek:
on a slight tangent here: @lucifer1004 remember that Nemo.jl is GPL licensed while Arblib.jl is MIT;
copying such small snippet shouldn't be a problem, but in general it's advisable not to look at GPL code when submitting to MIT licensed project.

src/Arblib.jl
lucifer1004: Updated

kalmarek (approved):

kalmarek:
@lucifer1004 Thanks for contributing!

PR #133, opened 2021-09-19
Fix scalar addition and subtraction for polynomials, missing special function and comparison with Rational

Joel-Dahne:
This fixes three things
1. Some bugs in the recent update to scalar addition and subtraction for polynomials related to normalization and handling of the zero polynomial.
2. A missing method from `SpecialFunctions` which is implemented by Arb but which I had missed the first time.
3. It fixes comparison between `Rational` and `Arb`. `Julia.Base` implements a special method for comparing `Rational` and `AbstractFloat` which doesn't work well for `Arb`. I explicitly defined these methods to just convert the rational to `Arb`.

I also bumps the version to `0.5.0`. Technically this release doesn't add anything which is not backwards compatible. However the new code for threading has some potential to lead to issues so to be conservative I though it would be better to make it 0.5.0 instead of 0.4.3.

### DIFF
--- a/src/poly.jl
+++ b/src/poly.jl
@@ -312,24 +312,35 @@ for (T, Tel, Tel_inplace) in [
 ]
     @eval function Base.:+(p::$T, c::$Tel_inplace)
         res = copy(p)
+        # Handle p being the zero polynomial
+        iszero(p) && return set_coeff!(res, 0, c)
         res0 = ref(res, 0)
         add!(res0, res0, c)
-        return res
+        return normalise!(res)
     end
     @eval Base.:+(p::$T, c::$Tel) = p + convert(eltype(p), c)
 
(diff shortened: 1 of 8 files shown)


src/poly.jl
```diff
@@ -312,24 +312,35 @@ for (T, Tel, Tel_inplace) in [
 ]
     @eval function Base.:+(p::$T, c::$Tel_inplace)
         res = copy(p)
+        # Handle p being the zero polynomial
+        iszero(p) && return set_coeff!(res, 0, c)
         res0 = ref(res, 0)
         add!(res0, res0, c)
-        return res
+        return normalise!(res)
```
kalmarek, line 319: Could you comment why `normalise!` here is necessary?

kalmarek (approved):

kalmarek:
Thanks!

Issue #134, opened 2021-09-21
Cannot convert BigInt to Arb


lucifer1004:
Currently, trying to do something like

```julia
a = Arb(big"3284738924733239284")
```

throws `InexactError`.

lucifer1004:
I ran into this issue when trying to use `Arblib.jl` with `WignerSymbols.jl`, in which the Clebsch-Gordan coefficients are stored as `RationalRoot{BigInt}`.

lucifer1004:
A possible workaround is to convert `BigInt` to `BigFloat` first, then convert the `BigFloat` to `Arb`.

Joel-Dahne:
I have also encountered this at some point and went with the same solution of converting to `BigInt` first. 

To properly handle this you would have to make a constructor which first converts the `BigInt` (which is just a GMP integer in the background) into a Flint integer (`fmpz_t`) and then construct an `Arb` from that. This would require creating a struct for storing a `fmpz_t` as well. 

When I last encountered this I didn't feel like putting all that work into it, but in the long run I definitely think it is something we should support. Feel free to submit a pull request if you want to get it fixed soon! Otherwise I'll probably try to get it done at some point at least!

Joel-Dahne:
Closed by #137

PR #136, opened 2021-09-22
Construct Arf/Arb/Arc from BigInt

lucifer1004:
solves #134

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -8,12 +8,14 @@ Arf(x; prec::Integer = _precision(x)) = set!(Arf(prec = prec), x)
 # disambiguation
 Arf(x::Arf; prec::Integer = precision(x)) = set!(Arf(prec = prec), x)
 Arf(x::Rational; prec::Integer = _precision(x)) = set!(Arf(prec = prec), x)
+Arf(x::BigInt; prec::Integer = precision(Arf)) = set_mpz!(Arf(prec = prec), x)
 
 #Arb
 Arb(x; prec::Integer = _precision(x)) = set!(Arb(prec = prec), x)
 # disambiguation
 Arb(x::Arb; prec::Integer = precision(x)) = set!(Arb(prec = prec), x)
 Arb(x::Rational; prec::Integer = _precision(x)) = set!(Arb(prec = prec), x)
+Arb(x::BigInt; prec::Integer = precision(Arb)) = set!(Arb(prec = prec), Arf(x, prec = prec))
 
 function Arb(str::AbstractString; prec::Integer = DEFAULT_PRECISION[])
     res = Arb(prec = prec)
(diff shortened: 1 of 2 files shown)


src/constructors.jl
```diff
@@ -8,12 +8,14 @@ Arf(x; prec::Integer = _precision(x)) = set!(Arf(prec = prec), x)
...
 Arf(x::Arf; prec::Integer = precision(x)) = set!(Arf(prec = prec), x)
 Arf(x::Rational; prec::Integer = _precision(x)) = set!(Arf(prec = prec), x)
+Arf(x::BigInt; prec::Integer = precision(Arf)) = set_mpz!(Arf(prec = prec), x)
 
 #Arb
 Arb(x; prec::Integer = _precision(x)) = set!(Arb(prec = prec), x)
 # disambiguation
 Arb(x::Arb; prec::Integer = precision(x)) = set!(Arb(prec = prec), x)
 Arb(x::Rational; prec::Integer = _precision(x)) = set!(Arb(prec = prec), x)
+Arb(x::BigInt; prec::Integer = precision(Arb)) = Arb(Arf(x; prec = prec); prec = prec)
```
kalmarek, line 18: I'd rather have to as
```julia
set!(Arb(prec = prec), Arf(x, prec = prec))
```
to keep in line of using `set!`-ters. I know this is not exactly the julia way, but that's what it is on the `arb` level

src/constructors.jl
```diff
@@ -32,6 +34,8 @@ Acb(
     im::Union{Real,arb_struct,arf_struct,Tuple{<:Real,<:Real}};
     prec::Integer = max(_precision(re), _precision(im)),
 ) = set!(Acb(prec = prec), re, im)
+Acb(re::BigInt, im::BigInt; prec::Integer = precision(Acb)) =
+    Acb(Arb(re; prec = prec), Arb(im; prec = prec); prec = prec)
```
kalmarek, line 38: ditto here: `set!(Acb(prec=prec), re, im)` would be preferred

kalmarek (changes requested):

kalmarek:
@lucifer1004 could you rebase on master?

lucifer1004:
Rebased.

Joel-Dahne:
I had completely missed that there is a `arf_set_mpz` method in Arb, that makes everything a lot easier since we don't have to go through `fmpz_t`! I think this will make things even easier actually, let me take a look at it!

Joel-Dahne:
Okay so I took a look at this! The main reason this has not been working properly is that the `_mpz` suffix of the method name had not been striped in the Julia name.  So what needs to be fixed is actually our parsing code. Since you might not be that familiar with it it might be easier if I do it. Let me put something together and make another pull request and see if we get it to work there!

PR #137, opened 2021-09-23
Add BigInt constructors

Joel-Dahne:
This adds constructors from `BigInt` and also removes the `_mpz` suffix from the Julia function names.

As I mentioned in #136 I realized that we had not been stripping the `_mpz` suffix of the functions names from Arb and setters for `BigInt` did therefore not work properly. In addition to that we have in many places used the `Integer` type when we really need something that can be converted to an `Int`, this made the code cleaner and made it work well with other small integer types. However it lead to problems for `BigInt` since they cannot always be converted to `Int`.

Notice that this would technically be a breaking change since some of the functions change name, e.g. `set_mpz!` now becomes `set!`. You could probably argue that the previous name was a bug though and I would surprised if anyone relies on it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "0.5.0"
+version = "0.5.1"
 
 [deps]
 Arb_jll = "d9960996-1013-53c9-9ba4-74a4155039c3"
--- a/src/arbcall.jl
+++ b/src/arbcall.jl
@@ -147,7 +147,7 @@ end
 function jlfname(
     arbfname,
     prefixes = ("arf", "arb", "acb", "mag", "mat", "vec", "poly", "scalar"),
-    suffixes = ("si", "ui", "d", "mag", "arf", "arb", "acb", "mpfr", "str");
+    suffixes = ("si", "ui", "d", "mag", "arf", "arb", "acb", "mpz", "mpfr", "str");
     inplace = false,
 )
     strs = filter(!isempty, split(arbfname, "_"))
@@ -170,7 +170,7 @@ end
 function jlfname(
     af::Arbfunction,
     prefixes = ("arf", "arb", "acb", "mag", "mat", "vec", "poly", "scalar"),
-    suffixes = ("si", "ui", "d", "mag", "arf", "arb", "acb", "mpfr", "str");
+    suffixes = ("si", "ui", "d", "mag", "arf", "arb", "acb", "mpz", "mpfr", "str");
     inplace = inplace(af),
 )
     return jlfname(arbfname(af), prefixes, suffixes, inplace = inplace)
--- a/src/setters.jl
+++ b/src/setters.jl
@@ -25,7 +25,11 @@ function set!(res::ArfLike, x::Rational; prec::Integer = precision(res))
     return res
 end
 
-function set!(res::ArfLike, x::Rational{BigInt}; prec::Integer = precision(res))
+function set!(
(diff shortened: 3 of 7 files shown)


Joel-Dahne:
I noticed that we don't actually have any tests for the `Arf` setters, I'll take a look at adding those tomorrow!

kalmarek (approved):
lgtm; thanks for adding the comment on disambiguation!

kalmarek:
while technically breaking one can argue that only the exported functions should be stable; anyway: version numbers come two a penny ;)

> In addition to that we have in many places used the Integer type when we really need something that can be converted to an Int, this made the code cleaner and made it work well with other small integer types. However it lead to problems for BigInt since they cannot always be converted to Int.

do you remember any particular cases where this convention might lead to problems?
it seems to me that functionality with `mpz_t` exists only for `arf`s:
```
$ grep -R " mpz_t" ./src/         
./src/arbcalls/arf.jl:arbcall"void arf_set_mpz(arf_t res, const mpz_t x)"
./src/arbcalls/arf.jl:arbcall"int arf_set_round_mpz(arf_t res, const mpz_t x, slong prec, arf_rnd_t rnd)"
./src/arbcalls/arf.jl:arbcall"int arf_mul_mpz(arf_t res, const arf_t x, const mpz_t y, slong prec, arf_rnd_t rnd)"
./src/arbcalls/arf.jl:arbcall"int arf_addmul_mpz(arf_t z, const arf_t x, const mpz_t y, slong prec, arf_rnd_t rnd)"
./src/arbcalls/arf.jl:arbcall"int arf_submul_mpz(arf_t z, const arf_t x, const mpz_t y, slong prec, arf_rnd_t rnd)"
```

Joel-Dahne:
If you use some integer type (other than the ones we have hard coded) with a value which is larger than `Int` you would currently get an error. Instead it might be natural to have it be converted to `BigInt` automatically. In the current version `Rational{UInt128}` and `Rational{Int128}` would not work, however this I think I will add.

Joel-Dahne:
Okay, I think I have covered everything we have discussed now! I'll leave it here until tomorrow in case you have some comments or I remember something more I should do.

kalmarek:
> If you use some integer type (other than the ones we have hard coded) with a value which is larger than `Int` you would currently get an error. Instead it might be natural to have it be converted to `BigInt` automatically. In the current version `Rational{UInt128}` and `Rational{Int128}` would not work, however this I think I will add.

hmm, this would have to be done during runtime I guess, by inserting comparison with `typemax(Int)`? or do you want to do it by adding those methods manually?

My main motivation to accept `Integer` whenever we have `Clong` in arbsignature was to provide working version for "small integers" _and_ throw when necessary (so that you don't get silent overflows:)

Joel-Dahne:
At the moment we don't support `fmpz_t` so it's not really a problem. The only thing we could do then is to convert to `Clong`, and if it throws an overflow error then so be it. If we at some point in the future start to support `fmpz` then we would have to think about how to handle it, but that's a problem for the future!

There are still a few cases where we have to be careful at the moment, construction and promotion. The construction part should be handled with this branch. What would not work fully is if someone uses an integer type which is not `UInt128`, `Int128` nor `BigInt` but which still overflows an `Int`, for example construction of `Arb` would not work in that case. I don't see this as a big problem, they will just have to convert to `BigInt` manually. Promotion should already work, this was the reason we defined
```julia
const _BitInteger = (Int == Int64 ? Base.BitInteger64 : Base.BitInteger32)
const _BitSigned = (Int == Int64 ? Base.BitSigned64 : Base.BitSigned32)
const _BitUnsigned = (Int == Int64 ? Base.BitUnsigned64 : Base.BitUnsigned32)
```
and created methods like
```julia
function Base.:+(x::_BitInteger, y::ArfOrRef)
    z = zero(y)
    add!(z, y, x)
    return z
end
```

Joel-Dahne:
I decided to only go with a increasing in the minor version in the end. Hopefully it doesn't skrew anyone. This is also our 100th closed pull request 🥳!

PR #141, opened 2021-11-06
Add deepcopy method for Arb structs

Joel-Dahne:
The other week I experienced weird behavior when running some computations. After a long time I managed to track it down to doing a `deepcopy` of some `Arb` values and we have never implemented a proper method for that. It used the default implementation in Julia which meant that the copy would share the same pointer to the limbs as the original. This commit adds proper methods for `deepcopy`.

### DIFF
--- a/src/arb_types.jl
+++ b/src/arb_types.jl
@@ -160,6 +160,7 @@ mutable struct acb_mat_struct
 end
 
 const ArbStructTypes = Union{
+    mag_struct,
     arf_struct,
     arb_struct,
     acb_struct,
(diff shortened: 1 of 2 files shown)


kalmarek (approved):
I was always under the impression that `stackdict` should be used only if structures are self-referrential? Could you explain it a bit more?

Other than this this seems just fine!

Joel-Dahne:
It also serves a purpose when a mutable structure occurs several times in a structure. For example
```julia
julia> value = [1, 2]
2-element Vector{Int64}:
 1
 2
julia> x = (value, value)
([1, 2], [1, 2])
julia> y = deepcopy(x)
([1, 2], [1, 2])
julia> x[1][1] = 5
5
julia> x
([5, 2], [5, 2])
julia> y
([1, 2], [1, 2])
julia> y[1][1] = 6
6
julia> y
([6, 2], [6, 2])
```
Since our structs are mutable and this is something that people can use I think it makes sense to preserve this behavior. Note that I actually added a specific test for the example above!

I have another PR in the pipeline so I'll wait with doing a release for now.

PR #142, opened 2021-11-18
Polynomial performance improvements

Joel-Dahne:
This adds a number of minor improvements to polynomials, mostly related to construction. 

The main goal is to improve the performance when constructing `ArbSeries` with of the form `ArbSeries([x, 1])`, mostly by reducing the number of allocations. It makes the following changes related to this
1. Allow passing tuples when constructing polynomials, saving one allocation in constructing the vector. So now `ArbSeries((x, 1))`  works. This change is breaking in the sense that previously passing a two-tuple `(a, b)` would set the first coefficient to `Arb((a, b))`, i.e. the interval from `a` to `b`. This could now be done with `ArbSeries(((x, 1),))`.
2. Make `setindex!` work well with `Int` using `arb_poly_set_coeff_si`. Previously this would convert the input to `Arb` before setting the coefficient.
3. Make a specialized constructor for two-tuples optimized for tuples with heterogeneous types. If you naively iterate over the elements in the tuple with a for loop the type inference doesn't work and you suffer in performance. By explicitly unfolding the loop type inference works well.
4. I have also removed the `fit_length!` in one place in the construction for series, this is already done by the underlying constructor so doesn't have to be done again.

Before the above changes the fastest way to construct an `ArbSeries` with coefficients `[x, 1]` was with `ArbSeries([x, one(x)])`, now it is `ArbSeries((x, 1))`, comparing the performance we see that the later is much faster and (more importantly) allocates a lot less.
```julia
julia> x = Arb(1)

julia> @btime ArbSeries([$x, one($x)])
  192.790 ns (4 allocations: 320 bytes)
1.0000000000000000000000000000000000000000000000000000000000000000000000000000 + 1.0000000000000000000000000000000000000000000000000000000000000000000000000000⋅x + 𝒪(x^2)

julia> @btime ArbSeries(($x, 1))
  112.588 ns (2 allocations: 144 bytes)
1.0000000000000000000000000000000000000000000000000000000000000000000000000000 + 1.0000000000000000000000000000000000000000000000000000000000000000000000000000⋅x + 𝒪(x^2)

```

In addition to the above I added `@inbounds` in one place where it was missing and better documentation for `Arblib.ref` for polynomials.

Since supporting constructing polynomials from tuples is a breaking change I have pushed the version to `0.6.0`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "0.5.2"
+version = "0.6.0"
 
 [deps]
 Arb_jll = "d9960996-1013-53c9-9ba4-74a4155039c3"
--- a/Project.toml
+++ b/Project.toml
@@ -14,5 +14,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 Arb_jll = "~200.2000"
 FLINT_jll = "~200.800"
-SpecialFunctions = "1.0"
+SpecialFunctions = "1.0, 2"
 julia = "1.3"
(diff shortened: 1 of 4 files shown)


kalmarek (approved):
This looks good.

Probably we should also support a vararg constructor as well, but this may wait for another time.

Joel-Dahne:
I took the liberty of adding the bump of`SpecialFunctions.jl` to this pull request from #143, it should not affect anything.

PR #146, opened 2021-12-21
Update Arb to version 2.21.1

### LINKED ISSUE #145: update arbcalls
Today I found
https://arblib.org/acb.html#c.acb_add_error_arb
which is not covered by our arbcalls.

@Joel-Dahne I remember you kept your private version of arb docs ;) could you generate new calls?

Joel-Dahne:
This closes #145. In the end handling the new formatting of the documentation was very simple, There were quite a lot of new functions added in 2.21.0 so this is a nice update to have.

Version 2.21.0 also adds the new floating point wrappers to Arb, https://www.arblib.org/arb_fpwrap.html. These would be nice to support at some point, but I'm leaving that for later.


Joel-Dahne:
I notice now that the new version of `Arb_jll` requires Julia at least version 1.6. I'm not sure what to do about this. Should we just drop support below Julia 1.6? Since 1.6 is the new LTS release it might be okay to drop anything below it?

kalmarek:
indeed it seems that this is the case; there seems to be a switch like this

https://github.com/JuliaPackaging/Yggdrasil/blob/5d62793f6c63dc9e90827e2d7e8623ab45ce5c6e/A/Arb/build_tarballs.jl#L26

that was set to `true` in my pull, so binaries only for `1.6` and later were built.

 
@fingolfin I know that everybody can push to Yggdrasil, but since You introduced the switch (https://github.com/JuliaPackaging/Yggdrasil/pull/3824): what is the release procedure? Should I set this to `false` and PR again?

kalmarek:
@Joel-Dahne I think it's best to drop support for julia <=1.6. The upstream (of Arblib_jll) doesn't care about supporting it and I don't want to spend my time on this either ;)

Joel-Dahne:
I agree! There are two options then, either we implicitly drop support for <= 1.6, basically we just remove the test for Julia 1.3 and let the requirement for `Arblib_jll` force the correct version, or we explicitly drop support for <= 1.6 by specifying it in the Project.toml. Explicitly dropping support has the benefit that we can start to use features from newer versions of Julia.

I would propose to only silently drop the support in this PR and then possibly follow up with another PR where we explicitly drop it. But I'm open to explicitly dropping it now as well. Either way I can try to make the required updates later today!

kalmarek:
@Joel-Dahne proceed as you find most convenient. dropping julia versions justifies minor version bump anyway

JeffreySarnoff:
fwiw, I would drop the 1.6 support as it is no longer the "standard" version.

Joel-Dahne:
Now I updated the CI and also the formatter workflow to use 1.6 as the oldest version. I also updated the version for JuliaFormatter.

fingolfin:
Hmm, the JLL supports Julia >= 1.6, so it's not clear to me why you want to drop support for Julia 1.6 (the LTS release) ? Maybe just a typo (but a repeated one...)


However, I'd also like to address a social point here: First off, I am sorry to not have reacted again to this, but on a typical day I get maybe a hundred GitHub notifications (so better email me directly or talk to me on Slack if you want to make sure I see something; all in all, I don't think pinging someone via an @-mention in a random PR/issue on a repository they have nothing to do with is a great way to communicate...

I would also like to point out that people tend to be on Christmas break during Christmas... (in my case, I was also ill), perhaps factor that in when asking stuff.

As it is, @kalmarek wrote:

> The upstream (of Arblib_jll) doesn't care about supporting it and I don't want to spend my time on this either ;)

I must say I found this quite offensive.

Who do you think is "the upstream (of Arblib_jll)" in your view, and what do you consider an appropriate way to get support from an open source project? Are 1-2 GitHub AT-mentions of one person (out of several) vaguely related, during Christmas holidays, and a lack of reactions, really sufficient ground to declare "they" (whoever that may, you leave it vague) "don't care" ? 

Also note that Yggdrasil is community driven, by all of us working together; JLLs don't have a fixed "maintainer" in general (perhaps they should have, but that's a hypothetical). I think more importantly is that most JLLs have a few primary "users" (= Julia packages depending on them), and it is indeed a good idea to coordinate changes to a JLL between all packages using that JLL (and for that matter, it would be good to record such a list of "packages using the JLL" in the build recipes, plus perhaps contact data for each package so one can contact these people when updating the JLL. However, that's not the same as being maintainer.

All in all, I hope `Arblib_jll` works well for you, and I am in general happy to help out people who try to do things related to Ygggdrasil / BinaryBuilder. I do not have any particular affiliation with Arb other than that I helped out with the JLL in the past; and I will update it again if projects or people I work with have need. But I am not its maintainer and will not handle general change requests to it.

And I hope we can leave out the snark and the underhand negative stuff in the future, that kind of thing helps nobody.

JeffreySarnoff:
@fingolfin speaking for myself (and maybe some others)
There is only respect for your efforts and good intent in their use.

With the most recent Arb C library  release, the author introduces a comprehensive set of mathematical functions designed to provide correctly rounded results working with args that are Float64s and Complex64s, and returning results that are  Float64s or Complex64s.
[arblib with Julia interop](https://arblib.org/arb_fpwrap.html).

Your `Arblib_jll` is well constructed and nicely organized, which simplifies doing more with Arb through Julia.

Joel-Dahne:
@fingolfin  The <= 1.6 is a typo indeed, we are definitely not dropping support for 1.6! Luckily the commit messages are correct at least :smiley: 

I'm very sorry if you got offended by anything that @kalmarek or I wrote! I completely understand that you might not have time to reply or might just miss the message for any reason and that is not a problem at all! Both Kalmarek and I have had times when we take weeks to reply so we are definitely not in a hurry here! And if we really needed your reply it would definitely have been our responsibility to contact you gain. Regarding Kalmareks comment, I read
> The upstream (of Arblib_jll) doesn't care about supporting it and I don't want to spend my time on this either ;)

more as a good motivation to drop support for <1.6, an indication that the Julia community is maybe moving towards 1.6 as the lowest supported version in some cases. In fact I read the "they" as "people that are much more knowledgeable in Julia than I am and obviously know more about this issue", so if they think dropping support for < 1.6 is reasonable than so should we. Though, given the rest of your comment I understand that there is no well defined "they" in this case. I wouldn't mind to continue to support earlier versions if anyone needed it, but from what it seems no one does and then there is no reason to bother with it. Dropping the supports only makes things easier for us in the end.

Regarding Yggdrasil I actually never knew how it worked before you mentioned it in #145. This is definitely my fault because I never bothered to look it up, but thank you for letting me know! I do agree that some coordination between packages using a certain jll might be helpful, I know that there are a couple of packages depending on `Arb_jll`.

Indeed `Arb_jll` has worked well for us. Probably best exemplified in that we up until now never had to think about it at all!

Thank you so much for your help!

kalmarek:
@fingolfin 
* of course `<1.6` was intended, we don't want to drop support for LTS (and it's fascinating to see that none of us actually noticed this, but silently we understood each other :).
* I'm very sorry if you got offended by my remark, this was absolutely not my intention! I just asked you since you were the last person to substantially edit the corresponding jll. I definitely didn't expect any support from you. In fact, since you didn't respond (for your own reasons, which are not my business :) I took the issue into my hands and submitted a PR to Yggdrasil. I absolutely *didn't expect* You to come and respond to my questions, in fact I'm rather *grateful* every time you eventually do!

> > The upstream (of Arblib_jll) doesn't care about supporting it and I don't want to spend my time on this either ;)

> I must say I found this quite offensive.

* I'm again sorry if you found this offensive. I wrote this after a quick discussion with Tommy (who reached to me in https://github.com/JuliaPackaging/Yggdrasil/pull/4090 ). In this (rather friendly, I think) conversation I learned that the switch I inquired about became irrelevant since 1.6 has become the LTS and it should be removed with the next pull. And because people from the Oscar project (whom I consider the primary user) prefer not to support <1.5 (e.g. they have more meaningful ways of spending their time) I felt we should follow suit. To clarify what I meant is 

> Since other people using `Arb_jll` decided not to support julia<1.6, and I can't invest my time in supporting it, I won't ask you or anybody else to invest their time in the issue.

I'm sorry that the remark sounded snarky to your ears. For me it was just an honest (if also a bit blunt) assessment of the situation.

* Finally addressing the social point: in the issue mentioned above I didn't ping you, neither here I ping Tommy, since I understand this is a minor issue not worth summoning you or Tommy. So I believe I do show consideration when trying to reach people (but feel free to ping him here if you think it's appropriate).

Joel-Dahne:
Oh, and @kalmarek can you approve the PR? :) Or do you have any remaining comments?

kalmarek (approved):

fingolfin:
All, esp. @kalmarek, thank you for the explanations. I now see that what @kalmarek was ambiguous, and I think I misread him. In any case, I retrospect I also think I overreacted, my apologies for that. No bad blood from my side.

PR #152, opened 2022-06-05
1.6 cleanup

Joel-Dahne:
Now that we require Julia >= 1.6 I though it would be reasonable to clean up some things that are not needed anymore. In the end I did a few more things as well:
1. I removed checks for version below 1.6
2. The interface for `rand` had some workarounds for a bug in Julia that are no longer needed. In particular it no longer uses `setprecision` which means it's thread safe.
3. I very much like using semicolons for keyword arguments. I have updated the code to use this everywhere, or at least everywhere I found. 
4. I updated the use of `_precision` slightly. I added a two argument version which is now used instead of the tuple version in general. So `_precision((a, b)) => _precision(a, b)`. I find this to be more natural.
5. I fixed a bug in the constructor of `ArbPoly` and related types which meant it would crash on `AbstractVector` which an indexing which is not one-based.
6. I renamed all test files to not end in `-test` (or `_test` in one case). I find the test-suffix unnecessary since the files are already in the test folder. This is similar to how for example `SpecialFunctions.jl` and `OffsetArrays.jl` does it.
7. Some more tiny fixes.
There should be no change in behavior, most of it is style changes. Some of the changes are more subjective, like using semicolon for keyword arguments and renaming the test files.


kalmarek (commented):
I didn't even knew that one can omit keywords after `;` ;) some calls become significantly shorter which is obviously a +; on the other hand it's slightly harder (for me) to read some of the code, so I don't really have strong opinion.

as for `_precision(t::Tuple)`: do we even need to keep it? it was an internal method so we can remove it altogether without introducing breaking changes, right?

Joel-Dahne:
Yeah the option to omit the keywords after `;` came in Julia v1.5. Personally I like it, in particular when most of the arguments are passed to another function, like in the `calc_integrate.jl` file. I can understand that it is a bit harder to read when you are not so used to it, but if you don't have a strong opinion I would say we keep the change.

kalmarek (approved):

kalmarek:
@Joel-Dahne merge at will, all looks good to me ;)

PR #155, opened 2022-07-27
Some new methods

Joel-Dahne:
I have added some new methods
- `SpecialFunctions.erfinv` and `SpecialFunctions.erfcinv`, these were added in a recent Arb release
- `setball(Arb, m, r)` which returns a ball with the given midpoint and radius. This was surprisingly awkward to do before. I have also exported this method.
- `add_error`, a non-mutating version of `Arblib.add_error!` which I have found myself using quite a bit. I have also exported this method.
Then I have made some very minor updates to comments and documentation.

I have some more updates in the pipeline after this as well :)

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -1,3 +1,5 @@
+export setball
+
 # Mag
 Mag(x) = set!(Mag(), x)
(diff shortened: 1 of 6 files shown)


kalmarek (approved):

kalmarek:
LGTM!

PR #158, opened 2022-08-24
Add hash and serialization methods

Joel-Dahne:
Up until now we had no special implementation of `hash` or `Serialization.serialize` for our types. In some cases this lead to undefined methods (e.g. `hash(::Arb)`), in other cases to incorrect results (e.g. `hash(::ArbPoly)`) and in some cases to segmentation faults (e.g. `serialize(::Arb)`). This PR contains proper implementations of `hash` and `serialize` for all our types.

I also fixed a bug for many constructors handling of empty arrays.

# Hash
This required quite a bit more work than I expected. The `hash` function should satisfy that if `isequal(x, y)` then `hash(x) == hash(y)`. In particular this should hold even when `x` and `y` are of different types numerical types but compare equal. This is implemented in Julia using the `Base.decompose` method, it doesn't have a doc-string but there are comments in the code describing its use. I have therefore implemented `Base.decompose` for `Mag` and `Arf`. For `Arb` and `Acb` we hash the underlying `Mag`s and `Arf`s, some care has to be taken to guarantee that hashes are equal when they should be.

For the non-`Number` types hashing is a bit easier and in many cases the default implementations are okay.

# Serialize
This makes use of `Arblib.dump_string` and `Arblib.load_string!` for the serialization. For containers is essentially iterates these on all the elements, with a little bit of extra work required for the series types. This works well but comes with a factor two overhead in size compared to an optimal implementation, the strings returned by `Arblib.dump_string` are given in base 16 and we are hence using only 4 of the 8 bits in each byte of the string. I think this overhead is well worth it compared to using a more complicated implementation.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install JuliaFormatter and format
         run: |
-          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="0.22.10"))'
+          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="1.0.9"))'
           julia  -e 'using JuliaFormatter; format(["./src", "./test"], verbose=true)'
       - name: Format check
         run: |
--- a/src/hash.jl
+++ b/src/hash.jl
@@ -1,0 +1,61 @@
+# The Base implementation of hash(::Real) is based on Base.decompose
+# to guarantee that all numbers that compare equal are given the same
+# hash. We implement Base.decompose for Mag and Arf.
+function Base.decompose(x::Union{mag_struct,Ptr{mag_struct}})::Tuple{UInt,BigInt,Int}
+    isinf(x) && return 1, 0, 0
+
+    if x isa Ptr{mag_struct}
+        x = unsafe_load(x)
+    end
+
+    pow = BigInt()
+    ccall(@libflint(fmpz_get_mpz), Nothing, (Ref{BigInt}, Ref{UInt}), pow, x.exponent)
+    pow -= 30
+
+    return x.mantissa, pow, 1
+end
+
+function Base.decompose(x::Union{arf_struct,Ptr{arf_struct}})::Tuple{BigInt,BigInt,Int}
+    isnan(x) && return 0, 0, 0
+    isinf(x) && return ifelse(x < 0, -1, 1), 0, 0
+
+    num = fmpz_struct()
+    pow = fmpz_struct()
+    ccall(
+        @libarb(arf_get_fmpz_2exp),
+        Cvoid,
+        (Ref{fmpz_struct}, Ref{fmpz_struct}, Ref{arf_struct}),
+        num,
+        pow,
+        x,
+    )
+
+    return BigInt(num), BigInt(pow), 1
+end
+
+Base.decompose(x::Union{MagOrRef,ArfOrRef}) = Base.decompose(cstruct(x))
+
+# Hashes of structs are computed using the method for the wrapping
+# type
+Base.hash(x::mag_struct, h::UInt) = hash(Mag(x), h)
+Base.hash(x::arf_struct, h::UInt) = hash(Arf(x), h)
+Base.hash(x::arb_struct, h::UInt) = hash(Arb(x), h)
+Base.hash(x::acb_struct, h::UInt) = hash(Acb(x), h)
+
+# Hashes of Mag and Arf are computed using the Base implementation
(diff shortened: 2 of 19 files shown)


kalmarek:
phew! that's quite a chunk! I'm afraid I will hardly have time to properly review this until 9-10 of September.

Joel-Dahne:
No worries! I'm using my local copy so it wont effect my work. Take all the time you need!

.github/workflows/format.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install JuliaFormatter and format
         run: |
-          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="0.22.10"))'
+          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="1.0.9"))'
```
kalmarek, line 21: this is a side note from this pr:

How about adapting a `JuliaFormatter` style for this project?
This works by placing a `.JuliaFormatter.toml` file in the main dir with e.g. content:
```toml
# Configuration file for JuliaFormatter.jl
# For more information, see: https://domluna.github.io/JuliaFormatter.jl/stable/config/

always_for_in = true
always_use_return = true
margin = 80
remove_extra_newlines = true
short_to_long_function_def = true
```

src/hash.jl
```diff
@@ -0,0 +1,80 @@
...
+# hash. We implement Base.decompose for Mag and Arf.
+function Base.decompose(x::Union{mag_struct,Ptr{mag_struct}})::Tuple{UInt,BigInt,Int}
+    isinf(x) && return 1, 0, 0
+
+    if x isa Ptr{mag_struct}
+        x = unsafe_load(x)
+    end
+
+    pow = BigInt()
+    ccall(@libflint(fmpz_get_mpz), Nothing, (Ref{BigInt}, Ref{UInt}), pow, x.exponent)
```
kalmarek, line 12: this is just a side-comment again: maybe we should consider replacing all the `ccall`s + our lib-macros by the new `@ccall` syntax? (if so I'll do it in a separate PR)
e.g. this would become
```julia
@ccall libflint.fmpz_get_mpz(pow::Ref{BigInt}, x.exponent::Ref{UInt})::Cvoid
```
if we replace `@libflint` by simple `const libflint = Arb_jll.libflint`

kalmarek, line 12: why `x.exponent` is `Ref`ed in the `ccall`?

src/hash.jl
```diff
@@ -0,0 +1,80 @@
...
+    pow -= 30
```
kalmarek, line 13: could you explain this "magic constant" (for posterity :)

src/hash.jl
```diff
@@ -0,0 +1,80 @@
+# The Base implementation of hash(::Real) is based on Base.decompose
+# to guarantee that all numbers that compare equal are given the same
+# hash. We implement Base.decompose for Mag and Arf.
+function Base.decompose(x::Union{mag_struct,Ptr{mag_struct}})::Tuple{UInt,BigInt,Int}
```
kalmarek, line 4: since `Base.decompose` is not documented could you write 2-3 lines describing what have you learned?

src/hash.jl
```diff
@@ -0,0 +1,80 @@
...
+end
+
+# arb_poly_struct and acb_poly_struct use default hash implementation,
+# this is okay since they don't implement an isequal method.
+function Base.hash(p::Union{ArbPoly,AcbPoly}, h::UInt)
+    for i = 0:degree(p)
+        h = hash(ref(p, i), h)
+    end
+    return h
+end
```
kalmarek, line 70: shouldn't we mix type of `p` into hash, e.g. by
```suggestion
function Base.hash(p::Union{ArbPoly,AcbPoly}, h::UInt)
    h = hash(typeof(p), h)
    for i = 0:degree(p)
        h = hash(ref(p, i), h)
    end
    return h
end
```

kalmarek (changes requested):
I don't know anything about serialization so I had just briefly looked at the code; nothing really sticks out there;

I have one question about `hash(p::Poly)` -- I think we should mix the type of `p` into `h`.

Other than this it looks good!

Joel-Dahne (commented):
Now it was my turn to not have a lot of time. I have finally taken a look at your comments and I think I have addressed all of them!

.github/workflows/format.yml
Joel-Dahne: Adding a style could maybe make sense. But would it not mainly be if we want to deviate from the default values? I haven't thought so much about the actual style that is used actually.

kalmarek (approved):

PR #159, opened 2022-10-16
Documentation

Joel-Dahne:
I have been thinking about improving the state of the documentation for a while now. This is still very much work in progress but at least a starting point. 

To run it locally you can follow the instructions from [Documenter.jl](https://juliadocs.github.io/Documenter.jl/stable/man/guide/). I do the following (I write it down mostly so I don't forget until next time): start a Julia REPL and activate the `Arblib.jl` repository, then run `include("docs/make.jl")`. I then run a local web server with `python3 -m http.server --bind localhost`.

It is the first time I create documentation for a Julia package and I have yet to figure out how to e.g. setup GitHub to build the documentation and such. 

Any comments or ideas for the documentation are appreciated! I just wanted to get something started to have something to build on.


kalmarek:
Hey, I've added the gh action for documentation; I'm not an expert though I just copied it from a different project that I created documentation for a long time ago ;)

src/ArbCall/ArbFunction.jl
```diff
@@ -16,6 +16,8 @@ declaration in C. Note that note all types of function declarations
 are supported, only those which are relevant for the Arb library.
 
 ```jldoctest
+julia> using Arblib
+
```
kalmarek, line 20: @Joel-Dahne It should be enough to add something like
```julia
DocMeta.setdocmeta!(
    Arblib,
    :DocTestSetup,
    :(using Arblib;);
    recursive = true,
)
```
to `docs/make.jl` so that in every `jldoctest` `Arblib` is available.
See https://documenter.juliadocs.org/stable/man/doctests/#Module-level-metadata

kalmarek: I'm documenting my own package at this very moment and I'm learning this as we speak ;)

Joel-Dahne: Great! I felt like there should be such a functionality but I couldn't immediately find it.

kalmarek:
> The main issue now is warnings for missing docs and cross references. Missing docs is not really a problem. Bit I haven't figure out how to handle the cross references. For example we reference `Base.cconvert` and it complains about not finding the docs for it. But it is a Base function so I don't know how to handle that.

It is not possible to link to a method outside of the documented package. See e.g
https://github.com/JuliaDocs/Documenter.jl/pull/1351
https://github.com/JuliaDocs/Documenter.jl/issues/1343

In this regard rust generated docs are so much better...

Joel-Dahne:
I have now removed some cross references and added the required documentation for some others. The only warnings remaining are for docs not included in the documentation, but these are mostly internal functions and I don't think it makes much sense to include them.

I also added documentation for the `XLike` types.

With this I think the documentation is "done". The most important parts are there, but it could certainly benefit from improvements at a later stage. The next step would be to actually deploy the documentation and link to it in the README. I'll see if I can figure out how that works!

Joel-Dahne:
Okay, I think this might be ready now! We'll only really know if it works until we deploy it though...

Joel-Dahne:
@kalmarek Could you approve this PR as well? Then I'll merge this and rebase the other one on this.

kalmarek (approved):

kalmarek:
@Joel-Dahne Sorry it took me so long; I'm traveling right now ;)

Joel-Dahne:
No worries at all! Lets hope that this works now!

PR #160, opened 2023-01-05
Remove constructor of Int from `ArfLike` and add from `ArfOrRef`

Joel-Dahne:
`ArfLike` includes `Ptr{Arblib.arf_struct}` and for pointers there is already a conversion to Int defined in base Julia which we should probably not override. 

I noticed this issue because I was looking for method invalidations using `SnoopCompile`. More precisely I ran
```julia
using SnoopCompile
invalidations = @snoopr begin
    # Your code goes here!
    using Arblib
end;
trees = SnoopCompile.invalidation_trees(invalidations)
```
which gave as output
```
6-element Vector{SnoopCompile.MethodInvalidations}:
 inserting Int64(x::ArfLike; rnd) in Arblib at /home/joeldahne/Syncthing/Work/Code/Arblib.jl/src/constructors.jl:76 invalidated:
   backedges: 1: superseding Int64(x::Ptr) in Core at boot.jl:774 with MethodInstance for Int64(::Ptr) (1 children)
   4 mt_cache

 inserting copyto!(dest::BitArray, bc::Base.Broadcast.Broadcasted{Nothing}) in Base.Broadcast at broadcast.jl:968 invalidated:
   mt_backedges: 1: signature copyto!(dest::BitArray, bc::Base.Broadcast.Broadcasted{Nothing}) in Base.Broadcast at broadcast.jl:968 (formerly copyto!(dest::AbstractArray, bc::Base.Broadcast.Broadcasted{Nothing}) in Base.Broadcast at broadcast.jl:948) triggered MethodInstance for copyto!(::BitVector, ::Base.Broadcast.Broadcasted{Nothing, Tuple{Base.OneTo{Int64}}, Type{Arblib.ArbCall.Carg}, Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(strip), Tuple{Vector{SubString{String}}}}}}) (5 children)

 inserting isnan(x::Union{Arb, ArbRef}) in Arblib at /home/joeldahne/Syncthing/Work/Code/Arblib.jl/src/predicates.jl:108 invalidated:
   backedges: 1: superseding isnan(x::AbstractFloat) in Base at float.jl:496 with MethodInstance for isnan(::AbstractFloat) (9 children)

 inserting deepcopy_internal(x::T, stackdict::IdDict) where T<:Union{Arblib.acb_mat_struct, Arblib.arb_mat_struct} in Arblib at /home/joeldahne/Syncthing/Work/Code/Arblib.jl/src/arb_types.jl:192 invalidated:
   backedges: 1: superseding deepcopy_internal(x, stackdict::IdDict) in Base at deepcopy.jl:53 with MethodInstance for Base.deepcopy_internal(::Any, ::IdDict{Any, Any}) (41 children)
   1 mt_cache

 inserting print(io::IO, s::Compat.Splat) in Compat at /home/joeldahne/.julia/packages/Compat/HCarO/src/Compat.jl:630 invalidated:
   backedges: 1: superseding print(io::IO, f::Function) in Base at show.jl:469 with MethodInstance for print(::IOBuffer, ::Function) (52 children)
   1 mt_cache

 inserting show(io::IO, ::Type{AcbMatrixLike}) in Arblib at /home/joeldahne/Syncthing/Work/Code/Arblib.jl/src/show.jl:132 invalidated:
   backedges: 1: superseding show(io::IO, x::Type) in Base at show.jl:881 with MethodInstance for show(::IOBuffer, ::Union) (1 children)
              2: superseding show(io::IO, x::Type) in Base at show.jl:881 with MethodInstance for show(::IOContext, ::Type) (11 children)
              3: superseding show(io::IO, x::Type) in Base at show.jl:881 with MethodInstance for show(::IOContext{Base.TTY}, ::Type) (31 children)
              4: superseding show(io::IO, x::Type) in Base at show.jl:881 with MethodInstance for show(::IOContext{IOBuffer}, ::Type) (95 children)
              5: superseding show(io::IO, x::Type) in Base at show.jl:881 with MethodInstance for show(::IOBuffer, ::Type) (1452 children)
   53 mt_cache
```
I noticed that we invalidated `Int64(x::Ptr)`, which seemed problematic. I don't think any of the other invalidations are problematic, though I don't fully understand the second one.

It could maybe be reasonable to also remove the constructors for `Float64(::Ptr{ArbExtras.mag_struct})`, `Float64(::Ptr{Arblib.arf_struct})` and `Float64(::Ptr{Arblib.arb_struct})` as well as `ComplexF64(::Ptr{Arblib.acb_struct})`. Though I don't think they would be problematic in any way so I'm fine with keeping them as well.


kalmarek:
Just to make it clear in my head: we remove this definition only because of invalidations (=longer waiting times), and not because of any type-piracy (and therefore silently wrong code)?

Joel-Dahne:
I don't think this method is used a lot so the invalidation is not that important, it just happened to be that I found the issue when looking for invalidation issues. 

The issue is that we overload a method with another one which has a different meaning. It is technically not type-piracy since we own the `Arblib.arb_struct` type.

Julia defines `Int(x::Ptr)  = bitcast(Int, x)` and this has the same meaning no matter the type of the pointer. Our method gives a different meaning to `Int(x::Ptr{Arblib.arb_struct})` which does not agree with the general one.

kalmarek (approved):

PR #163, opened 2023-08-03
Make setindex! always use set!

Joel-Dahne:
Makes it so that `res[] = x` is a shorthand for `Arblib.set!(res, x)` in all cases, previously it was only implemented for some `x` types. In particular this makes `ArbVector([(1, 2)])` work as expected.

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -45,7 +45,6 @@ Base.@propagate_inbounds function Base.setindex!(
     i::Integer,
     j::Integer,
 )
-    @boundscheck checkbounds(A, i, j)
     ref(A, i, j)[] = x
     return x
 end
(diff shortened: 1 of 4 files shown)


src/matrix.jl
```diff
@@ -45,7 +45,6 @@ Base.@propagate_inbounds function Base.setindex!(
     i::Integer,
     j::Integer,
 )
-    @boundscheck checkbounds(A, i, j)
```
kalmarek, line 48: hmm, why removing those checkbounds? are we checking those in `ref` as well?

kalmarek (approved):

src/matrix.jl
Joel-Dahne: Exactly, the `ref` method already checks bounds for both `ArbVector` and `ArbMatrix`.

Joel-Dahne: I only happened to notice them because I went through all occurrences of `setindex!`.

PR #166, opened 2023-10-09
Poly multiarg methods an optimizations

Joel-Dahne:
This implements the multi-argument addition and multiplication also for polynomials. I also noticed that `isnan` and `isfinite` for polynomials allocate a lot as written, so I added an optimization for that.

This week I'm attending a [Flint workshop](https://flintlib.github.io/workshop2023.html) so I will try to get as much work as possible done. You of course might not have the time to review everything, so I'll try to make successive PRs and then we can merge them one by one at some later stage.


Joel-Dahne:
I should mention that this is based on the #159 PR, only the last two commits are new. I don't think there is a good way to mark this PR as being based on another PR and only show the differences unfortunately...

src/predicates.jl
```diff
@@ -108,10 +108,18 @@ end
 Base.isnan(x::ArbOrRef) = isnan(midref(x))
 Base.isnan(x::AcbOrRef) = isnan(midref(realref(x))) || isnan(midref(imagref(x)))
 
-Base.isnan(p::Union{ArbPoly,ArbSeries,AcbPoly,AcbSeries}) =
-    any(isnan, @inbounds p[i] for i = 0:degree(p))
-Base.isfinite(p::Union{ArbPoly,ArbSeries,AcbPoly,AcbSeries}) =
-    all(isfinite, @inbounds p[i] for i = 0:degree(p))
+function Base.isnan(p::Union{ArbPoly,ArbSeries,AcbPoly,AcbSeries})
+    @inbounds for i = 0:degree(cstruct(p))
```
kalmarek, line 112: hmm, why is this `cstruct` here? is `degree(p)` not defined/gives different result?
(simple in-line comment would clarify this)

docs/src/index.md
```diff
@@ -0,0 +1,20 @@
...
+[ArbNumerics.jl](https://github.com/JeffreySarnoff/ArbNumerics.jl).
+
+The **goal** of Arblib.jl is to supply a **low lever wrapper** of the
+methods in Arb as well as a **high level interface**. The low level
+wrapper should allow for writing methods using mutability and with
+performance very close to that of those written in C. The high level
+interface should make it easy to use in generic Julia code, similarly
+to how `BigFloat` is a wrapper around the MPFR library. In addition it
+should be possible to seamlessly switch between the high level
+interface and the low level wrapper when needed.
```
kalmarek, line 15: hehe, this **high level interface** is not something we had in mind when we started, but why not, if you have time and motivation ;)

docs/src/interface-mutable.md
```diff
@@ -0,0 +1,85 @@
...
+
+eval(p, x) = begin
+    res = zero(x)
+    xi = one(x)
+    for i in 0:Arblib.degree(p)
+        res += p[i] * xi
+        xi *= x
+    end
+    return res
+end
```
kalmarek, line 70: why defining like this instead of `function eval(...)`?

src/ArbCall/ArbFPWrapFunction.jl
```diff
@@ -171,6 +177,22 @@ function jlcode(af::ArbFPWrapFunction, jl_fname = jlfname(af))
...
+generate the code for a corresponding method with
+[`Arblib.ArbCall.jlcode`](@ref) and evaluate the code.
+
+For example
+```
+arbfpwrapcall"int arb_fpwrap_double_exp(double * res, double x, int flags)"
+```
+defines the method
+```
+fpwrap_exp(x::Union{Float16, Float32, Float64}; error_on_failure::Bool = Arblib.ArbCall.fpwrap_error_on_failure_default(), correct_rounding::Bool = false, work_limit::Integer = 8)
```
kalmarek, line 193: just whitespace to distinguish args from kwargs:
```suggestion
fpwrap_exp(x::Union{Float16, Float32, Float64};
    error_on_failure::Bool = Arblib.ArbCall.fpwrap_error_on_failure_default(),
    correct_rounding::Bool = false, work_limit::Integer = 8
)
```

kalmarek (approved):

kalmarek:
@Joel-Dahne I think I've gone through most of this before and I see nothing contentious ;)
I made a few comments/suggestions, would be nice if you could apply/respond to them, but these are really minor.

Other than this this is ready to go.

Kudos for tackling the documentation!

docs/src/index.md
Joel-Dahne: With high level interface I only really mean that we overload the standard Julia functions like `*` and `sin`, so that you don't have to only work directly with the low-level methods like `Arblib.mul!` and `Arblib.sin!`. And this we already do!

src/ArbCall/ArbFPWrapFunction.jl
Joel-Dahne: Fixed!

src/predicates.jl
Joel-Dahne: Should have added a comment from the beginning! It's a microoptimization that avoids iterating over coefficients known to be zero for series. It would be more natural to write `degree(p.poly)` for series, but then that doesn't work for polynomials. Added a comment now!

PR #168, opened 2023-10-12
Avoid use of `XLike`

Joel-Dahne:
This PR removes the use of `XLike` types in several places. This leads to breaking changes since the methods are no longer defined for structs and pointers. This is in line with making the high level interface more focused on the high level types. It depends on #167.

The most notable change is the change to `isequal` (and `==`) for structs. They no longer compare equal to other types, so an `arb_struct` is never equal to an `Arb` even if the underlying struct is the same. This fixes earlier problems with `isequal` not being transitive. Before we had
```julia
julia> isequal(Mag().mag, Mag())
true

julia> isequal(Mag(), Arb())
true

julia> isequal(Arb(), Arb().arb)
true

julia> isequal(Mag().mag, Arb().arb)
false
```
whereas we now have
```julia
julia> isequal(Mag().mag, Mag())
false

julia> isequal(Mag(), Arb())
true

julia> isequal(Arb(), Arb().arb)
false

julia> isequal(Mag().mag, Arb().arb)
false
```
With this change the `hash` methods for structs have also been reworked. They no longer give the same hash as for the non-struct types.

Other changes are reduced usage of `XLike`  in some constructors a few other methods. Some argument types have also been slightly simplified in the process.

### DIFF
--- a/src/hash.jl
+++ b/src/hash.jl
@@ -47,45 +47,21 @@ end
 Base.decompose(x::Union{MagOrRef,ArfOrRef}) = Base.decompose(cstruct(x))
 
 # Hashes of structs are computed using the method for the wrapping
-# type and then hashed together with a constant. Compare with
-# Base.h_imag
-if UInt === UInt64
-    const h_mag = 0x5d9b5fdb71940a8e
-    const h_arf = 0x9587d52d253e1dd1
-    const h_arb = 0x5209056683bc3f7a
-    const h_acb = 0x9ca782e3b04fab90
-    const h_arb_vec = 0x25e99ee61f9a8792
-    const h_acb_vec = 0x05f9e7ef121a65fe
-    const h_arb_poly = 0xd821aefd7c0bb7d4
-    const h_acb_poly = 0xed00c6f74947b740
-    const h_arb_mat = 0x1caddefdf435e388
-    const h_acb_mat = 0x68e0fd60c8bdd5d1
-else
-    const h_mag = 0x8fa1c5cc
-    const h_arf = 0x695f4554
-    const h_arb = 0x769d5541
-    const h_acb = 0x5eda1191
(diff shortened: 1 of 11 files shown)


src/hash.jl
```diff
@@ -47,16 +47,50 @@ end
...
+Base.hash(x::arb_vec_struct, h::UInt) =
+    hash(ArbVector(x, shallow = true), hash(h_arb_vec, h))
+Base.hash(x::acb_vec_struct, h::UInt) =
+    hash(AcbVector(x, shallow = true), hash(h_acb_vec, h))
+Base.hash(x::arb_poly_struct, h::UInt) = hash(ArbPoly(x), hash(h_arb_poly, h))
+Base.hash(x::acb_poly_struct, h::UInt) = hash(AcbPoly(x), hash(h_acb_poly, h))
+Base.hash(x::arb_mat_struct, h::UInt) =
+    hash(ArbMatrix(x, shallow = true), hash(h_arb_mat, h))
+Base.hash(x::acb_mat_struct, h::UInt) =
+    hash(AcbMatrix(x, shallow = true), hash(h_acb_mat, h))
```
kalmarek, line 88: @Joel-Dahne I suppose you computed those hashes by hash(...some_type...);
What I always do is to
```julia
Base.hash(x::mag_struct, h::UInt) = hash(Mag(x), hash(hash(typeof(x)), h))
```
I was always convinced that constant propagation should be enough.
Did you try this approach?
I don't particularly like the pollution of the namespace by all of those `h_...`, but if the above idea doesn't cut it that'll have to stay. Or we could maybe turn this into a `let` statement?

kalmarek (changes requested):
I like the direction of these changes (i.e. to replace the use of bare structs by `Ref`-types. 

Could have a look at hash comment?

kalmarek (approved):

Issue #165, opened 2023-09-25
Towards v1.0


Joel-Dahne:
I have been thinking that it might be time for a 1.0 release. The interface has been fairly stable for some time and I see no reason to stay on `0.x`. There are a couple of things I would like to tackle before 1.0
- Finish the transition to Flint v3
- Finish the first version of the documentation
- Reconsider the overloading of some `Base` methods, in particular `union` and `intersect` but also `contains`
- Reduce usage of `XLike` types in favor for `XOrRef` types in the high level interface

The first two points are more or less self explanatory. For the documentation I have made some progress the last month or so.

For `union` and `intersect` I have realized that the way we define them contradict the version in `Base`, and is also easily misused. The issue is that `Real` types are so both `union` and `intersect` already have a well defined meaning for those types, which is different from how we currently use it. For example
```julia
julia> using Arblib
julia> union(1.0, 2.0)
2-element Vector{Float64}:
 1.0
 2.0
julia> union(Arb(1), 2.0)
2-element Vector{Arb}:
 1.0000000000000000000000000000000000000000000000000000000000000000000000000000
 2.0000000000000000000000000000000000000000000000000000000000000000000000000000
julia> union(Arb(1), Arb(2))
[+/- 2.01]
julia> intersect(1.0, 1.0)
1-element Vector{Float64}:
 1.0
julia> intersect(Arb(1.0), 1.0)
1-element Vector{Arb}:
 1.0000000000000000000000000000000000000000000000000000000000000000000000000000
julia> intersect(Arb(1.0), Arb(1.0))
1.0000000000000000000000000000000000000000000000000000000000000000000000000000
```
For that reason I think it would be better to use some other name for taking the union or intersection of two `Arb` or `Acb`. My proposal would be `setunion` and `setintersection`, in line with for example `getinterval`, `getball` and `setball`. But I'm open to other ideas! The method `Base.contains` also has similar issues, but I haven't thought so much about how to deal with that one.

Regarding `XLike` types I have noticed that I very rarely need to use anything other than the `XOrRef`  types with the high-level interface. When I do use the, it is mainly together with the lower-level interface.  In most cases the high-level interface already uses `XOrRef`, but there are some exceptions:
- Most predicates use `XLike`
- `hash`
- a couple more spread out though the code
[...]

PR #174, opened 2023-11-06
Release 1.0

Joel-Dahne:
Except for one minor thing, see below, I think we are now ready for a 1.0 release. Everything I mention in #165 has been taken care of. This PR contains three changes
- Update the [compat requirements for standard libraries](https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958)
- Remove some comments about early development in documentation
- Set the version to 1.0

The one remaining thing I had in mind was if we should add a doi identifier for the repository, using e.g. [Zenodo](https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content). I have not done this before and I don't know exactly how to think about this. So far I have mentioned Arblib.jl in three separate publications and it seems like it would make sense to have a doi link combined with proper archiving. Do you have any thoughts about this?

Also, do you have anything else you think we should handle before 1.0?

Once we do a release I thought I would send an email about it on the Flint email list and possibly also a post at the Julia discourse. 

Sorry for keeping you so busy 😅

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "0.8.1"
+version = "1.0"
 
 [deps]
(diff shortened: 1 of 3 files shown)


kalmarek:
@Joel-Dahne I've done this already a few times with Zenodo, I'll take care of this.

Follow up: so this was just flipping one switch there ;) after a new release we'll get the doi and a badge to boast about it ;D

(those guys at Zenodo are simply amazing when it comes to ease of use!).

kalmarek (approved):

PR #182, opened 2024-02-26
Flint 3.1.0

Joel-Dahne:
Flint 3.1.0 was recently released. The Julia wrapper is not yet updated, but hopefully some takes care of that being fixed. Once that is the case I'll update the Flint compat as well.

I have reparsed the documentation at the 3.1.0 tag in Flint, which also adds the `acb_theta` module.


Joel-Dahne:
This should be ready for merging now! The nightly build fails due to an update to `setprecision(f, T)` for Julia what will be 1.12, something to take care of at some point!

kalmarek (approved):

PR #184, opened 2024-04-14
Fix admonitions

devmotion:
When browsing the docs I noticed that admonitions are not displayed correctly. This PR should fix the problem.

Ref: https://documenter.juliadocs.org/stable/showcase/#Admonitions


Joel-Dahne:
Thank you David, this (and the other two PRs) is great! I'm handing in the final version of my PhD thesis now on Friday so I wont have much time to take a look at it before then. Feel free to take care of it if you want @kalmarek, otherwise I'll take a look after Friday!

devmotion:
Good luck, I hope these last days are not too stressful 🙂

kalmarek (approved):

PR #185, opened 2024-04-14
Add dependabot and update CI

devmotion:
I noticed that a few GH actions are outdated. The PR adds dependabot to make it easier to keep them updated. Additionally, CI is updated a bit.


kalmarek:
@devmotion Cool, and thanks! I always find keeping up with the bots/actions too tiresome. I'm very glad you had a look and decided to fix it!

kalmarek (approved):

kalmarek (approved):

PR #186, opened 2024-04-14
Add Aqua tests and fix unbound type parameters

devmotion:
The PR adds [Aqua tests](https://juliatesting.github.io/Aqua.jl/dev/). On master, both the tests for method ambiguities and for unbound type parameters are broken. I fixed the unbound type parameters in this PR but the method ambiguity issues are not resolved.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,6 +11,7 @@ Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 
 [compat]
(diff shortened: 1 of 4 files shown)


kalmarek:
@devmotion thanks for those tests using Aqua! the reported overwrites seem to be real and we need to find them

devmotion:
It would be good to fix these but I thought it would be better to go through them in separate PRs. So I intentionally did not include any fixes in this PR.

Joel-Dahne:
Yes, I guess we can merge this now then!

Joel-Dahne (approved):

Issue #171, opened 2023-10-25
Possibly use ScopedValues for precision (and rounding?)


Joel-Dahne:
There is a (currently open) [PR](https://github.com/JuliaLang/julia/pull/51362) for Julia about using `ScopedValues` for `setprecision` with `BigFloat`. If this ends up being merged it seems like a good idea for us to do the same. I could see this being useful in a number of situations.

PR #187, opened 2024-04-16
Fix `setprecision` errors on Julia nightly

devmotion:
https://github.com/JuliaLang/julia/pull/51362 (which switched to ScopeValue for MPFR precision and rounding which, however, seems to cause performance regressions; see also #171) of restricted `Base.setprecision(f::Function, ::Type, ::Integer; kwargs...)` to `Type{BigFloat}` which broke tests of Arblib on Julia nightly. The PR copies the implementation for `ArbTypes`.

### DIFF
--- a/src/precision.jl
+++ b/src/precision.jl
@@ -1,51 +1,40 @@
 const DEFAULT_PRECISION = Ref{Int}(256)
 
-if VERSION < v"1.8.0"
-    # Types
-    Base.precision(::Type{<:Union{ArbTypes,ArbStructTypes,Ptr{<:ArbStructTypes}}}) =
-        DEFAULT_PRECISION[]
-    # Types not storing their precision
-    Base.precision(x::Union{ArbStructTypes,Ptr{<:ArbStructTypes}}) = DEFAULT_PRECISION[]
-    # Types storing their precision
-    Base.precision(x::ArbTypes) = x.prec
-    # Mag doesn't store a precision
-    Base.precision(::MagOrRef) = DEFAULT_PRECISION[]
-    # ArbSeries and AcbSeries don't store their precision directly
-    Base.precision(x::Union{ArbSeries,AcbSeries}) = precision(x.poly)
-else
-    # Since Julia 1.8.0 Base.precision calls Base._precision and by
-    # overloading that we automatically support giving base argument.
-
-    # Types
-    if VERSION < v"1.11.0-DEV"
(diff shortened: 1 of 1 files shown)


kalmarek (approved):

kalmarek:
@devmotion thanks especially for including those links to PRs in base and julia dev revisions.
What is the easiest way to find them in the future?

devmotion:
> What is the easiest way to find them in the future?

The process is actually quite simple: You go to the PR in the Julia repo that introduced the change, obtain the merge commit (easiest by clicking on the link to the commit shown on Github when the PR was merged, eg above here it would be "commit f0efb0f" shown in the line "kalmarek merged ..."), and then obtain the corresponding Julia version number with the contrib/commit-name.sh script in the Julia repo. From the Compat.jl README:

> If you're adding additional compatibility code to this package, the [contrib/commit-name.sh](https://github.com/JuliaLang/julia/blob/master/contrib/commit-name.sh) script in the base Julia repository is useful for extracting the version number from a git commit SHA. For example, from the git repository of julia, run something like this:
>
> ```
> bash $ contrib/commit-name.sh a378b60fe483130d0d30206deb8ba662e93944da
> 0.5.0-dev+2023
> ```
>
> This prints a version number corresponding to the specified commit of the form X.Y.Z-aaa+NNNN, and you can then test whether Julia is at least this version by VERSION >= v"X.Y.Z-aaa+NNNN".

nsajko:
FTR, it's much better to report an issue to upstream as a bug report, rather than working around it: JuliaLang/julia#55899

src/precision.jl
```diff
@@ -1,51 +1,40 @@
...
+    elseif VERSION < v"1.11.0-DEV.1363"
+        # Since Julia 1.8.0 Base.precision calls Base._precision and by
+        # overloading that we automatically support giving base argument.
+        # Ref: https://github.com/JuliaLang/julia/pull/42428
+        :_precision
+    else
+        # Since Julia 1.11.0, the single-argument `Base._precision` is
+        # renamed to `Base._precision_with_base_2`
+        # Ref: https://github.com/JuliaLang/julia/pull/52910
+        :_precision_with_base_2
```
nsajko, line 15: The name is prefixed with an underscore to make it clear it's internal. Why trespass!?

kalmarek: Trespassing or not ;)

I personally don't have time now to delve into (internals or not) the design of `Base.precision` to figure out what would suit us best and will be least maintenance in the future. If you do have a clear solution we're all ears ;)

PR #191, opened 2024-04-17
Fix codecov integration

devmotion:
#188 broke codecov integration: Version 4  of the action does not support tokenless iploads anymore. Thus currently codecov uploads will always fail, see eg https://github.com/kalmarek/Arblib.jl/actions/runs/8724232667/job/23934347761#step:8:34. This error was masked by the `fail_ci_if_error: false`, so you might want to consider changing it to `true` (at the cost of spurious test failures sometimes since codecov servers are not completely reliable in my experience).

To fix codecov integration, in addition to this PR you have to go to the codecov webpage, navigate to the settings for this repo, and copy the token shown there to a Github repo secret named `CODECOV_TOKEN`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,4 @@ jobs:
           flags: unittests
           name: codecov-umbrella
           fail_ci_if_error: false
+          token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,5 +40,5 @@ jobs:
           file: ./lcov.info
           flags: unittests
           name: codecov-umbrella
-          fail_ci_if_error: false
+          fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}


kalmarek:
@devmotion Thanks for detailed info and instructions!

I've added the token to the repo. I'd personally change `fail_ci_if_error` to `true`, but let @Joel-Dahne voice his opinion on this;)

Joel-Dahne:
Changing `fail_ci_if_error` to `true` sounds reasonable to me as well!

.github/workflows/ci.yml
```diff
@@ -41,3 +41,4 @@ jobs:
           flags: unittests
           name: codecov-umbrella
           fail_ci_if_error: false
```
devmotion, line 43:
```suggestion
          fail_ci_if_error: true
```

kalmarek:
@devmotion and now macOS fails as it tries to use (the non-existent) token from your fork? the other ones just resolve to tokenless upload. Is this ok to be merged?

devmotion:
An alternative to `true` could be something like https://github.com/JuliaStats/StatsBase.jl/pull/920 (which was motivated by test failures in PRs from forks that we saw in StatsBase).

kalmarek:
I'm fine with simple `true`; the more complicated it is the more maintenance burden it becomes

kalmarek (approved):

PR #198, opened 2025-01-14
Improve low level interface for series

Joel-Dahne:
Flint has a number of functions for computing truncated series expansions, these generally have a name ending in `_series`. This PR adds special handling of the low level wrappers for these functions, something I have wanted for a long time.

In addition to the low level methods that were previously constructed for these functions we now add another method. The new method removes the `_series` suffix from the method name, restricts the input type to only series types (and not polynomial types) and takes the default length of the computed series from the first argument. It also handles `arb_poly_mullow` as a special case, since this is morally the same as the function `arb_poly_mul_series` (same for `acb_poly`).

The main motivation for the new methods is to make it easier to write generic code using mutability. For example now `Arblib.sin!` can be applied to both `Arb` and `ArbSeries`, before one had to use `Arblib.sin!` for `Arb` and `Arblib.sin_series!` for `ArbSeries`. As an example I have in some [old project](https://github.com/Joel-Dahne/HighestCuspedWave.jl/blob/2d6dd92e7a4fe7d995dda6684116224d331730fb/src/BurgersHilbert/evaluation.jl#L559C1-L572C8) code like this
```julia
w!(res::Arb, x::Arb) = begin
    Arblib.inv!(res, x)
    Arblib.add!(res, res, 1)
    Arblib.log!(res, res)
    return Arblib.sqrt!(res, res)
end
w!(res::ArbSeries, x::ArbSeries) = begin
    len = length(x)
    Arblib.inv_series!(res, x, len)
    Arblib.add!(res, res, 1)
    Arblib.log_series!(res, res, len)
    return Arblib.sqrt_series!(res, res, len)
end
```
with this update it would be possible to just do
```julia
w!(res::T, x::T) where {T<:Union{Arb,ArbSeries}} = begin
    Arblib.inv!(res, x)
    Arblib.add!(res, res, 1)
    Arblib.log!(res, res)
    return Arblib.sqrt!(res, res)
end
```

I have added information about all of this to the documentation. In the future we should probably try to expand the section about mutable arithmetic. Quite a lot of the work in Arblib.jl is for making mutable arithmetic convenient, and I think the documentation on how to use this in a good way could be improved. To get the documentation to compile locally I had to add `Arblib` as a dependency to it, not sure what has changed since before (hopefully the build on Github works).

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,4 +1,5 @@
 [deps]
+Arblib = "fb37089c-8514-4489-9461-98f9c8763369"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
--- a/docs/src/interface-mutable.md
+++ b/docs/src/interface-mutable.md
@@ -1,6 +1,10 @@
 # Mutable arithmetic
 The high level interface can be combined with the low level wrapper to
-allow for efficient computations using mutable arithmetic.
+allow for efficient computations using mutable arithmetic. Making use
+of mutable arithmetic can in some cases have a significant impact on
+the performance of the code, in particular for multithreaded code
+(where avoiding to run the GC is generally more important) and code
+running many iterations of simple computations.
 
 In the future it would be nice to have an interface to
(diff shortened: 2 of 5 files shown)


kalmarek (approved):

kalmarek:
While I've never worked with Arb series this looks good to me!

PR #199, opened 2025-01-20
Add new type `Acf`

Joel-Dahne:
This PR is a little bit of a mess. The goal is to add support for the relatively new Flint [acf](https://flintlib.org/doc/acf.html) Flint type. There is not actually a lot of functionality implemented for the type in Flint, so it is maybe a bit unnecessary to wrap it. But at least it is one step forward in getting full coverage of the real and complex number types in Flint.

Adding a completely new type requires touching more or less every part of the Arblib.jl code base, so the change log for this PR is huge. As I said it is currently a little bit of a mess. Feel free to ignore the PR for now, I'll most likely make some more adjustments.


fredrik-johansson:
Worth noting that there is a decent amount of functionality implemented via generic rings, e.g. e.g. ``gr_exp()`` should work with ``acf``.

Joel-Dahne:
Indeed! Currently there is no (convenient) way of accessing the functionality of the generic interface in Arblib.jl. I have thought a little bit about how to handle that, but have not gotten to the point of testing it. But on of the motivations for wrapping the `acf` type is indeed that it could be more useful once it can be used through the generic interface!

Joel-Dahne:
Okay, let me give some more comments on this now!

As mentioned the goal is to add the type `Acf` for wrapping the `afc` type in Flint. As you can see from the diff this amounts to adding and updating quite a lot of code in a lot of different places (the PR touches 44 files). Many of these changes are trivial albeit tedious, and it is easy to make copy and paste errors. The PR has a 100% test coverage, so that should at least give us some confidence that things are correct (of course the tests could also suffer from the same copy and paste issues...). Let me comment a bit on some of the more important parts.

1. The `acf` module in Flint is very limited (as Fredrik mentions the type is intended to be used through the generic interface, which is a completely different beast to tackle). This means that the type has only very basic functionality. It also means that we have to do some things a bit differently than for `Acb` since not functionality is not available. For example we have `Base.real(z::AcbOrRef) = get_real!(Arb(prec = _precision(z)), z)`, but `Base.real(z::AcfOrRef) = Arf(realref(z))`, since there is no `acf_get_real` function in Flint (in this case the alternative implementation is actually shorted, but that is besides the point).
2. I have defined the type `AcfRef`, but there is currently no native way of constructing this type (except using a raw pointer). I thought it would be a good idea to define it anyway since that makes the code look a lot more similar to the already existing code for the other types and in the future it might be useful.
3. The documentation for `Acf` is a lot better than for the other types. I might try to improve the documentation for the other types at some point...
4. I made some changes which are not strictly related to adding the `Acf` type, but which did make it easier to adjust the code for it. See for example the first commit in the PR. Going through the whole code base is a good way of finding issues!

I also have one design decision to ask about. There is currently no `acf_div` function, only an `acf_approx_div`. I'm not sure if we want to define `/` to use the approximate division or not? Currently I have not done that, `/` is simply not defined for `Acf`. My motivation for this was that both `arf` and `acf` are really intended for doing correct rounding and the user could expect that the functions are correctly rounded. In Flint the approximate versions are clearly marked as such, but `/` would not be. The counterargument is that correct rounding is extremely brittle and Julia is very happy to compute for example `Arf(5)^6` using power by squaring, which of course will not be correctly rounded. So already as it is one cannot really trust the `Arf` or `Acf` types to be correctly rounded, so it would then be perfectly fine to define `/` to use the approximate division instead. I'm currently leaning towards letting '/' use the approximate version (so not how it is currently implemented), but I would be happy to hear your thoughts.

kalmarek:
I'd go for the approximated div, just to make things easier on the developer side (deduplication etc). Correct rounding is generally not taken for granted in julia when using floating point types. If `acf_div` becomes available later in the future we could alter this in the future as a "bugfix" or bringing the intended behaviour.

Many thanks for putting this work!

Joel-Dahne:
I added the division using the approximate version now. With this the PR should be ready to merge.

kalmarek (approved):

PR #201, opened 2025-01-27
Include theta module

Joel-Dahne:
I noticed that we generate the file `src/arbcalls/acb_theta.jl` but we don't actually include it in the end, this PR fixes that. Note that it is built on top of #199, so we should wait with merging it until that one is handled.


Joel-Dahne:
I don't think there is any rush at all in getting this merged, so it can definitely wait!

Joel-Dahne:
Now this should be mergable!

kalmarek (approved):

PR #205, opened 2025-03-20
Set version to 1.3.0

Joel-Dahne:
I think this is a reasonable time to release 1.3.0. The `nfloat` wrapping will probably be at least another month, so there is no point in waiting for that. For reference, the new additions since last release are:
- Improvements to low-level series interface #198 
- Wrapping of `Acf` #199
- Update to Fling 3.2.0 #204


kalmarek (approved):

kalmarek:
@Joel-Dahne Great, thanks for working on this! merge and release at will.

Issue #206, opened 2025-03-25
creating ranges of `Arb` fails


simeonschaub:
e.g.

```julia-repl
julia> Arb(0):Arb(10)
ERROR: rem not defined for Arb
Stacktrace:
  [1] error(::String, ::String, ::Type)
    @ Base ./error.jl:44
  [2] no_op_err(name::String, T::Type)
    @ Base ./promotion.jl:624
  [3] rem(x::Arb, y::Arb)
    @ Base ./promotion.jl:643
  [4] mod(x::Arb, y::Arb)
    @ Base ./float.jl:605
  [5] rem(x::Arb, y::Arb, ::RoundingMode{:Down})
    @ Base ./div.jl:100
  [6] div(x::Arb, y::Arb, r::RoundingMode{:Down})
    @ Base ./div.jl:372
  [7] fld(a::Arb, b::Arb)
    @ Base ./div.jl:321
  [8] _colon(::Base.Ordered, ::Base.ArithmeticRounds, start::Arb, step::Arb, stop::Arb)
    @ Base ./range.jl:26
  [9] (::Colon)(start::Arb, step::Arb, stop::Arb)
    @ Base ./range.jl:20
 [10] (::Colon)(a::Arb, b::Arb)
    @ Base ./range.jl:14
 [11] top-level scope
    @ REPL[8]:1
```

This works for other floating point types. `Arb` should probably implement `fld`, maybe something like `floor!(Arb(), div!(Arb, x, y))`?

Joel-Dahne:
Indeed, there are a lot of missing functionality when it comes to rounding of `Arb` values. It seems like what is needed in this case to to define `div(x::Arb, y::Arb, r::RoundingMode)` (which is then automatically used by `fld`) and also `Integer(x::Arb)`. 

 I made a (draft) PR #207 implementing these two (as well as `round`). With this it is possible to do
```julia
julia> Arb(1):Arb(10)
1.0:1.0:10.0
```
Note that due to the way rounding works for `Arb` this can lead to issues in some cases. For example we have
```julia
julia> Arb(π):Arb(π)+1
ERROR: InexactError: BigInt([+/- 1.01])
Stacktrace:
 [1] BigInt(x::Arb)
   @ Arblib ~/Syncthing/Work/Code/Arblib.jl/src/conversion.jl:96
 [2] Integer
   @ ~/Syncthing/Work/Code/Arblib.jl/src/conversion.jl:109 [inlined]
 [3] convert
[...]

PR #207, opened 2025-03-26
Implement rounding related methods

Joel-Dahne:
This implements `round` for `Arf`, `Acf`, `Arb` and `Acb` as well as conversion to general integer types (not only `Int` and `BigInt`). It also implements `div` for `Arb`. In theory one could implement `div` for `Arf`, but getting that to be correctly rounded (which the `div` function does specify) seems like it could be a little bit of work and I therefore left it out.

The motivation for this is discussed in #206. 

There is still some work to be done, for example there are currently not any tests...


Joel-Dahne:
It seems like there is some work with getting the ambiguity tests to pass as well...

Joel-Dahne:
I finally got around to finishing this. I simplified some of the implementations and added tests for everything. I also added a test to make sure that `Arb(1):Arb(10)` actually works, since this was the initial motivation for this PR.

When implementing the tests for `RoundNearest` I ran into some issues, that turned out to be due to a [bug in Flint](https://github.com/flintlib/flint/issues/2293). The bug has now been fixed in Flint, but while we wait for a new release I added a temporary implementation which is correct but not optimal (it is a fixed version of the original implementation in Flint, the fixed version in Flint uses a slightly different approach).

Some of the tests are however failing on Julia v1.6. It seems like some of the handling of rounding has been added in later version, and hence are not supported on 1.6. For example `round(1.5, RoundFromZero)` doesn't work on 1.6 and the definitions
```julia
trunc(::Type{T}, x) where T = round(T, x, RoundToZero)
floor(::Type{T}, x) where T = round(T, x, RoundDown)
 ceil(::Type{T}, x) where T = round(T, x, RoundUp)
```
also seem to not exist in 1.6, leading to `floor(Int, Arb(1.5))` not working. Working around all of this would probably take some effort. Maybe this is a good time to drop support for 1.6 in favor of 1.10 (which is the current LTS)?

Joel-Dahne:
I rebased this on top of #211 to make sure that everything works as expected on nightly. So this would now have to be rebased once #211 is merged.

kalmarek (approved):

kalmarek:
@Joel-Dahne Thanks for all of those changes and staying up to date with julia-nightly ;)

kalmarek:
@Joel-Dahne minor release?

Joel-Dahne:
I'm currently looking at making `setprecision` use `ScopeValue`. I think making a release once that is finished (hopefully soon) would be a good idea!

PR #209, opened 2025-05-17
Set minimum Julia version to 1.10

Joel-Dahne:
I was busy with other things, but finally got time to get back to this!

The motivation for dropping support for versions older than Julia 1.10 is #207. The implementation I made doesn't work on Julia 1.6 due to some rounding related functions not existing. For this specific PR the issues could be worked around, but there are some potential future features where this would not be possible. The main thing I have in mind is #171, switching `setprecision` to use `ScopedValue`. The Base implementation of `ScopedValue` was introduced in Julia 1.11 but there is a compatible implementation in [ScopedValues.jl](https://github.com/vchuravy/ScopedValues.jl), that package does however require Julia 1.8. So to implement that feature we would be required to drop support for Julia 1.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - '1.10'
           - '1'
           - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -19,7 +19,7 @@ Random = "1.6"
 Serialization = "1.6"
 SpecialFunctions = "1, 2"
 Test = "1.6"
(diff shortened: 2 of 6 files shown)


Joel-Dahne:
The failure in nightly is due to a recent change to a [recent change to `minimum` and `maximum`](https://github.com/JuliaLang/julia/pull/58267) and is unrelated to the changes here. That is something to fix in a separate PR! The change on nightly is in fact related to [an issue](https://github.com/JuliaLang/julia/issues/45932#event-17561993167) I created a couple of yours ago with issues I found when getting `minimum` and `maximum` to work for `Arb`.

kalmarek (approved):

PR #212, opened 2025-05-19
Make `setprecision` use `ScopedValue`

### LINKED ISSUE #171: Possibly use ScopedValues for precision (and rounding?)

Joel-Dahne:
The main purpose of this PR is to make `setprecision` use `ScopedValue` to make it possible to change default precision inside a dynamic scope instead of only globally. This makes it possible to use it in a thread-safe way. Note that `Base.ScopedValues` was only introduced in v1.11, it therefore uses [ScopedValues.jl](https://github.com/vchuravy/ScopedValues.jl) to be compatible with v1.10 (for v1.11 and later this package is just a wrapper of the Base implementation). It is a bit difficult to properly test the thread-safety of the new version, I made an attempt using `Threads.@spawn` for which the test does fail without this PR.

I also took the opportunity to slightly restructure the implementation of `precision` and `setprecision`. It no longer uses any internal functions of `Base` and should hence be a bit more stable with respect to future changes to Julia.

Finally I also added a section of the documentation that goes a bit more in depth about how the precision is used and how it differs from `BigFloat`. It is definitely not perfect, but probably suffices as a starting point.

I also set the version to 1.4.0, so we can make a minor release after merging this!

This closes #171


kalmarek (approved):

kalmarek:
@Joel-Dahne This is as clean as usual ❤️!

Joel-Dahne:
I can make a release later today!

kalmarek:
That's the easy part of the release process ;D

Issue #217, opened 2025-07-26
possible bug in computation of inverse


lucaferranti:
Maybe I just didn't RTFM properly, but is the following correct?

```julia
julia> a = Arb("0.0000000001", prec=3)
[+/- 1.03e-10]

julia> 1 / a
[+/- 1.49e+10]
```

shouldn't it be `[+/- inf]` ?

lucaferranti:
ah, I think I was just mislead by the printing

```julia
julia> lbound(a)
6e-11
```

I understood the `[+/- 1.03e10]` as `[0 +/- 1.03e10]` , but it seems it's not the case. Is this printing intentional?

Joel-Dahne:
This would be a case of RTFM, if we had actually written that part of the manual... As it is now it is a case of look through old closed issues and find related ones, which is certainly not something I would expect everyone to do.

The issue is indeed that the printing is misleading, see the discussions in #208 for more background. The value `[+/- 1.03e10]` should indeed be understood as `[0 +/- 1.03e10]`, but that is not a tight enclosure of `a`. So even though the printed enclosure contains zero the actual enclosure does in this case not include zero. 

Some more details for this case:

You can get the midpoint and radius of `a` (which is how it is internally represented) with
```julia
julia> a = Arb("0.0000000001", prec=3)
[+/- 1.03e-10]

julia> getball(a)
(9e-11, 1.47641917e-11)
```
and here we can clearly see that the radius (the second value) is smaller than the midpoint (the first value), so the ball indeed doesn't cross zero. This is however still a bit misleading due to the extremely low precision used in this case, which the printing is not tuned for. We can get the exact values by explicitly printing more digits using the (undocumented) `digits` argument to `string`.
```julia
julia> string.(getball(a), digits = 100)
("8.731149137020111083984375e-11", "1.476419174092392927466477203779504634439945220947265625e-11")
```
In this case `digits = 100` is enough to get the exact values, but this depends on the values in question.

lucaferranti:
Thanks for the reply! 

In this case, I think printing something like

```julia
9e-11 +/- 1.47641917e-11
```

for the `MIME"text/plain"` (nice-printing for repl) would make sense.  (bonus points for using the unicode `±` ;) )

what do you think?

kalmarek:
[...]

Issue #208, opened 2025-04-14
Bug in show function


PBrdng:
Hi, 

we discovered a weird bug when displaying Intervals from `Arblib`. Some intervals seem to be printed incorrectly. See here:

https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/issues/635

The flag `more = true` seems to fix the bug. Do you know what's going on here and how to fix it?

kalmarek:
@PBrdng can you give an example of what interval is printed incorrectly and what do you think it should print?

PBrdng:
Of course. The example interval is an output from `HomotopyContinuation.jl`:

```julia
using HomotopyContinuation

@var x y;
F = System([x^2 + 1e-25 * y^2, y - 1]);

candidates = [
    [-3.399609423161143e-28 + 3.1622776601683797e-13im, 1.0 + 0.0im],
    [4.0326928841388684e-28 - 3.1622776601683797e-13im, 1.0 + 0.0im],
]

cert_result = certify(F, candidates; threading = false)
list_of_certificates = certificates(cert_result)
intervals = certified_solution_interval.(list_of_certificates)
```

Here, `intervals` is a vector of intervals:
```julia
julia> intervals
2-element Vector{Arblib.AcbMatrix}:
 [[+/- 8.18e-14] + [+/- 3.98e-13]im; [1.00000000000000 +/- 4.45e-16] + [+/- 5.05e-29]im;;]
 [[+/- 8.18e-14] + [+/- 3.98e-13]im; [1.00000000000000 +/- 4.45e-16] + [+/- 5.05e-29]im;;]
```

However, this is not displayed correctly. The correct answer is 
```julia
julia> for i in 1:2
    for j in 1:2
        s = string(intervals[i][j]; digits = 16, more = true)
        println(s)
    end
end
[+/- 8.18e-14] + [3.162277660168379e-13 +/- 8.18e-14]im
[1.000000000000000 +/- 4.45e-16] + [+/- 5.05e-29]im
[+/- 8.18e-14] + [-3.162277660168379e-13 +/- 8.18e-14]im
[1.000000000000000 +/- 4.45e-16] + [+/- 5.05e-29]im
```

You can also see that the first print is incorrect, because the printed intervals intersect in `[0; 1]`. However:
```julia
julia> isdisjoint(intervals[1], intervals[2])
true
```

Is this example enough or should I set up an example without `HomotopyContinuation.jl`?

[...]

Issue #84, opened 2020-10-31
default show for Arb might be confusing


kalmarek:
This is an interval containing `0`:
```julia
julia> let x = sin(Arb(π))
           @info "" x contains(x, 0) Arblib.midref(x) Float64(Arblib.radref(x))
       end
┌ Info: 
│   x = [+/- 2.83e-77]
│   contains(x, 0) = true
│   Arblib.midref(x) = 1.0969174409793520767073235159733943661910882381141135060401716459409529154108e-77
└   Float64(Arblib.radref(x)) = 1.730607217577946e-77
```

This is an interval containing `1`:
```julia
julia> let x = sin(Arb(π)/2)
           @info "sin(Arb(π))" x contains(x, 1) Arblib.midref(x) Float64(Arblib.radref(x))
       end
┌ Info: sin(Arb(π))
│   x = [1.000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.74e-77]
│   contains(x, 1) = true
│   Arblib.midref(x) = 0.99999999999999999999999999999999999999999999999999999999999999999999999999999
└   Float64(Arblib.radref(x)) = 8.737900781332737e-78
```
This is an interval containing `1.5`:
```julia
julia> let x = 1.5sin(Arb(π)/2)
           @info "1.5sin(Arb(π)/2)" x contains(2x, 3) Arblib.midref(x) Float64(Arblib.radref(x))
       end
┌ Info: 1.5sin(Arb(π)/2)
│   x = [1.5000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 4.77e-77]
│   contains(2x, 3) = true
│   Arblib.midref(x) = 1.5
└   Float64(Arblib.radref(x)) = 3.0379188354575523e-77
```
let's try the same with `set_interval!`:

```julia
julia> let x = Arblib.set_interval!(Arb(), Arf(-0.5), Arf(0.5))
           @info "[-0.5, 0.5]" x contains(x, 0) Arblib.midref(x) Float64(Arblib.radref(x))
       end
┌ Info: [-0.5, 0.5]
│   x = [+/- 0.501]
│   contains(x, 0) = true
│   Arblib.midref(x) = 0
└   Float64(Arblib.radref(x)) = 0.5000000009313226

julia> let x = Arblib.set_interval!(Arb(), Arf(0.5), Arf(1.5))
[...]

PR #218, opened 2025-07-28
Add documentation, primarily related to printing

Joel-Dahne:
As apparent from the various issues, e.g. #84, #208 and #217, the printing of `Arb` values can be very confusing. This PR doesn't change any of this behavior (though it does change the minimum number of digits printed to 1), but adds some documentation related to this. It adds:
1. Documentation to the `string` function that is used for controlled printing of `Arb` values.
2. Documentation to `dump_string` and `load_string`.
3. A section "Printing" to the online documentation, discussing issues with printing.  It's a collection of various gotchas that I have encountered myself over the years.
4. Documentation to `Mag`, `Arf`, `Arb` and `Acb`. This is not really related to the printing, but I added a warning to the documentation of `Arb` related to the printing. The hope is that user would be more likely to see it if it is part of the types documentation and not only under the `string` function.

As always the documentation could be improved further, but it is hopefully better than the old situation of not having anything. Let me know if you see anything that should be clarified or that could be added!


src/show.jl
```diff
@@ -71,6 +71,95 @@ function Base.string(
...
+  to display.
+- `more::Bool = false`: If `true`, display more (possibly incorrect)
+  digits.
+- `no_radius::Bool = false`: If `true`, display the radius is not
+  included in the output. Unless `more` is set, the output is rounded
+  so that the midpoint is correct to 1 ulp. As a special case, if
+  there are no significant digits after rounding, the result will be
+  shown as ``0e+n``, meaning that the result is between ``-1e+n`` and
+  ``1e+n`` (following the contract that the output is correct to
+  within one unit in the only shown digit).
```
kalmarek, line 122:
```suggestion
- `no_radius::Bool = false`: If `true`, the radius is not
  displayed in the output. Unless `more` is set, the output is rounded
  so that the midpoint is correct to 1 ulp. As a special case, if
  there are no significant digits after rounding, the result will be
  shown as ``0e+n``, meaning that the result is between ``-1e+n`` and
  ``1e+n`` (following the contract that the output is correct to
  within one unit in the only shown digit).
```

src/show.jl
```diff
@@ -71,6 +71,95 @@ function Base.string(
...
+- `condense::Integer = 0`: If non-zero, strings of more than three
+  times `condense` consecutive digits are condensed, only printing the
+  leading and trailing `condense` digits along with brackets
+  indicating the number of digits omitted (useful when computing
+  values to extremely high precision).
+- `unicode::Bool = false`: If `true`, use unicode characters for parts
+  of the output in the output, e.g., `±` instead of `+/-`.
```
kalmarek, line 129:
```suggestion
- `unicode::Bool = false`: If `true`, use unicode characters in the output, e.g., `±` instead of `+/-`.
```

Joel-Dahne:
The failures on Nightly seem to be related to Aqua.jl and not something we are responsible for fixing.

src/show.jl
```diff
@@ -71,6 +71,96 @@ function Base.string(
...
+  ``0e+n``, meaning that the result is between ``-1e+n`` and ``1e+n``
+  (following the contract that the output is correct to within one
+  unit in the only shown digit).
+- `condense::Integer = 0`: If non-zero, strings of more than three
+  times `condense` consecutive digits are condensed, only printing the
+  leading and trailing `condense` digits along with brackets
+  indicating the number of digits omitted (useful when computing
+  values to extremely high precision).
+- `unicode::Bool = false`: If `true`, use unicode characters in the
+  output in the output, e.g., `±` instead of `+/-`.
```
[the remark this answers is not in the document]
Joel-Dahne: Should be fixed now!

PR #235, opened 2026-04-28
Flint 3.5

Joel-Dahne:
Update to Flint 3.5.0, which was released last week.


Joel-Dahne:
Do you have some time to take a quick look at this @kalmarek ?

kalmarek (approved):

kalmarek:
@Joel-Dahne now that I'm working as swe I reduce my engagement in additional sw-related projects, so feel free also to proceed without me ;)

Joel-Dahne:
That is reasonable!

Currently I believe the repository is set up to require approval from at least one maintainer other than the one creating the PR. In case you won't be as active we could maybe disable that setting?

PR #238, opened 2026-08-20
Some fixes based on JET.jl report

Joel-Dahne:
I ran the `report_package` function from [JET.jl](https://github.com/aviatesk/JET.jl). It have a lot of false positives (mostly related to pessimistic union splitting), but also surfaced some real issues that I fixed. While fixing them I made some other minor adjustments as well. None of the fixes are particularly important, most of them are related to the low level struct types not implementing some functionality. The most relevant fixes are:

- Define `length` for vector structs and `size(A, d)` for matrix structs. Also update the eigenvalue functions to use length for vectors. While updating the eigenvalue functions I also made some minor adjustments to error messages and a bit of cleanup.
- I added `Nothing` as a possible parent type for all Ref types. Previously only `AcbRef` and `AcfRef` supported it. What made me add it to the other ones as well was that `parentstruct` was previously not defined for pointers to structs. This was a problem since e.g. `ArbLike` is given by `Union{Arb,ArbRef,cstructtype(Arb),Ptr{cstructtype(Arb)}}` and hence all functions using `parentstruct` and accepting `ArbLike` would not work for `Ptr{cstructtype(Arb)}`. In practice this is probably not very common and we basically do not have any tests for this. Nevertheless it felt like a reasonable fix to do.
- Some other minor fixes that should not be visible to users.

I set the version to 1.8.1 and will make a patch release with the fixes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arblib"
 uuid = "fb37089c-8514-4489-9461-98f9c8763369"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Sascha Timme <Sascha Timme <timme@math.tu-berlin.de>", "Joel Dahne <joel@dahne.eu>"]
-version = "1.8.0"
+version = "1.8.1"
 
 [deps]
 FLINT_jll = "e134572f-a0d5-539d-bddf-3cad8db41a82"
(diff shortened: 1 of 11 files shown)


Joel-Dahne:
I noticed a couple of issues in `src/eigen.jl` with return types and checking the flag returned by Flint. I fixed those and cleaned up the tests a bit. I also added some more tests for the integration, mostly because I wanted to improve the test coverage a bit.

PR #239, opened 2026-08-24
Add high level interface for FFT

Joel-Dahne:
This adds a high level interface to the FFT code of Flint. It depends on the currently unreleased version of Flint, more precisely the changes in [#2774](https://github.com/flintlib/flint/pull/2774), so the tests will not pass here. It currently includes a temporary commit related to reparsing the Flint documentation, once the next version of Flint is released this should be regenerated.

Note that Flint refers to FFTs as DFT. We here follow the Julia standard of referring to them as FFT. This could possibly be mentioned in the documentation.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -16,6 +16,7 @@ makedocs(
             "Types" => "interface-types.md",
             "Ball methods" => "interface-ball.md",
             "Integration" => "interface-integration.md",
+            "FFT" => "interface-fft.md",
             "Series" => "interface-series.md",
             "Mutable arithmetic" => "interface-mutable.md",
             "Precision" => "precision.md",
--- a/docs/src/interface-fft.md
+++ b/docs/src/interface-fft.md
@@ -1,0 +1,7 @@
+``` @docs
+Arblib.fft
+Arblib.ifft
+Arblib.fft!
+Arblib.ifft!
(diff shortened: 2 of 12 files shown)


Joel-Dahne:
This is also the first time I wrap any of the generic ring code in Flint. Though it is a very minor part here, only the struct and the error codes. It is however a good first step towards wrapping more of it!

fredrik-johansson:
> Note that Flint refers to FFTs as DFT

I'm open to changing this before we release a new version

Joel-Dahne:
> > Note that Flint refers to FFTs as DFT
> 
> I'm open to changing this before we release a new version

I don't have a strong opinion on this. DFT is technically the more correct name and given the more algebraic nature of Flint (compared to a general FFT package) I think using DFT is fitting. On the other hand I could definitely see some users completely missing the FFT support because they don't recognize the name.

For the Arblib wrapping code renaming it to `fft` in Flint would actually make the wrapping more complicated. Now `Arblib.dft!` is the low-level Flint function whereas `Arblib.fft!` is the higher level wrapper. If it was renamed to `fft` in Flint then I would have to figure out how to distinguish between the high and low versions of `Arblib.fft!`. This is a general drawback with the low-level interface and the high-level interface sharing the same namespace in Arblib.

